PR #5, opened 2022-08-25
 Add support for the new Realtime API including Broadcast and Presence

jan-tennert:
The new realtime api is now available. The update changes and adds a few things:
- Everything is now a channel
- A channel has a unique id
- Multiple clients can connect to the same channel with the same id
- You can send broadcasts and every client within the same channel will receive this broadcast (ignoring the database step and send it directly to other clients)
- You can still listen for database changes
- Presence (see example)

[Realtime blog post ](https://supabase.com/blog/supabase-realtime-multiplayer-general-availability)

Currently the new syntax looks like this:
```kotlin
val channel = supabaseClient.realtime.createChannel("#random") { //create channel
    //config
}
```
#### Listen for Postgres Changes
```kotlin
val changeFlow: Flow<PostgresAction.Insert> = channel.postgresChangeFlow<PostgresAction.Insert> {
     schema = "public"
     table = "test"
}
```
#### Broadcast messages to other clients
```kotlin
val broadcastFlow: Flow<Position> = channel.broadcastFlow("position") //under the event "position"

channel.broadcast("position", Position(20, 30)) //broadcast your positon to other clients (in the event "position")
```
#### Manage Presences
```kotlin
@Serializable
data class PresenceData(val userId: Int, val username: String)

channel.presenceFlow()
    .onEach {
        //joins and leaves are a map. The keys are either an id generated by realtime or a custom key which can be customized in the client builder. 
        otherUsers += joins.values.mapNotNull { it.stateAsOrNull<PresenceData>() }
        otherUsers -= leaves.values.mapNotNull { it.stateAsOrNull<PresenceData>() }
    }
    .launchIn(SomeScope)
channel.join()
channel.track(PresenceData(2, "Example")) //send your "state"/presence to other clients

### CODE REVIEW

**Comment by jan-tennert** [comment]
This pr will be frozen for the next two weeks because I'm not at home. But you can still leave some suggestions

**Comment by brezinajn** [comment]
How do you feel about adding support for flows to make the consumption more Kotlin idiomatic? In practice all it needs is just wrapping the callbacks in a CallbackFlow and provide a nice API. Wanted to open a PR, but now it looks it'd be better to wait for this update to merge.

**Comment by jan-tennert** [comment]
For the Realtime API?

**Comment by jan-tennert** [comment]
I mean yes I could do that directly in this pr. What syntax are you proposing?

**Comment by jan-tennert** [comment]
Something like this?
```kotlin
val channel = client.realtime.createChannel("id") {
    //config
}

val changeFlow = channel.postgresChangeFlow {
    schema etc
} //possibly with generics

val broadcastFlow = channel.broadcastFlow("event")

channel.join()
```
not sure about the function names

brezinajn:
Yes that's exactly what I had in mind. I played with an idea of adding generic param to PostgresAction for convenient deserialization.

jan-tennert:
yea maybe

jan-tennert:
@brezinajn I have a question about the flows: The callback flow works fine for the broadcasts but is almost always results to an exception for postgrest changes because you actually need to tell realtime which changes you want to receive. Because the callback for the changes is only added when the flow gets executed(?), realtime doesn't know of the requested changes. So something like this doesn't work ("always"):

```kotlin
val channel = supabaseClient.realtime.createChannel("#random") {

}
channel.postgrestChangeFlow<PostgresAction.Update>() {
    schema = "public"
    table = "test"
    event = "UPDATE"
}
    .onEach { println(it.record) }
    .launchIn(scope)
channel.join()
```

jan-tennert:
Now you could just separate the callbacks and the join config for the postgres changes, but that raises another problem: When you send the join payload you can match your registered postgres changes and the ones the server received. They also give you an id to identify which callbacks to call when receiving a change. You can't ignore that because on the actual change payload you get your record or old record, the event type (UDPATE, etc.), the schema, and the table but not the filter which you can also set (you get that on the event which mentioned the id, to update the ids locally).

jan-tennert:
Ok now it should work. The join payload and the callbacks are now independent

brezinajn:
> @brezinajn I have a question about the flows: The callback flow works fine for the broadcasts but is almost always results to an exception for postgrest changes because you actually need to tell realtime which changes you want to receive. Because the callback for the changes is only added when the flow gets executed or launched(?), realtime doesn't know of the requested changes. So something like this doesn't work ("always"):
> 
> ```kotlin
> val channel = supabaseClient.realtime.createChannel("#random") {
> 
> }
> channel.postgrestChangeFlow<PostgresAction.Update>() {
>     schema = "public"
>     table = "test"
>     event = "UPDATE"
> }
>     .onEach { println(it.record) }
>     .launchIn(scope)
> channel.join()
> ```
> 
> See
> 
> https://github.com/jan-tennert/SupaCompose/blob/ca3a1eab432ecf9be6c40f93a4aa572e50ee814e/Realtime/src/commonMain/kotlin/io/github/jan/supacompose/realtime/RealtimeChannel.kt#L199

I have a very basic implementation where I justy wrap onAll in a callbackFlow and filter the events later. I can see the use case for actually receiving only a subset of events. In that case I'd provide a set of event types to receive when creating the flow.

jan-tennert:
I updated the PR description with the new syntax. As I said in the new api realtime you send the server which changes you want to listen to (before you could just listen to one) Your solution worked before because you actually specify what you want to listen to before the listeners. (In the channel topic). The "topic" is now the channel id and you say which changes you want to listen to in the join payload. But it should work now because the callbacks and the actual join payload are now independent. It works now like this: When you call a function like postgresChangeFlow it first adds the filter you specify within the builder to the join payload and then adds the callback.

brezinajn:
Looks great. It'd be nice to have automatic type inference based on event type.
``` Kotlin
val changeFlow: Flow<PostgresAction.Insert> = channel.postgresChangeFlow {
     schema = "public"
     table = "test"
     event = EventType.Insert
}

val changeFlow: Flow<PostgresAction.Update> = channel.postgresChangeFlow {
     schema = "public"
     table = "test"
     event = EventType.Update
}

val changeFlow: Flow<PostgresAction.Delete> = channel.postgresChangeFlow {
     schema = "public"
     table = "test"
     event = EventType.Delete
}

val changeFlow: Flow<PostgresAction> = channel.postgresChangeFlow {
     schema = "public"
     table = "test"
     event = EventType.All
}

```

This should be possible to achieve by having
``` Kotlin
sealed class EventType<T: PostgresAction>{
    object All: EventType<PostgresAction>()
    object Insert: EventType<PostgresAction.Insert>()
    object Update: EventType<PostgresAction.Update>()
    object Delete: EventType<PostgresAction.Delete>()
}
```

jan-tennert:
The type inference did work, but its still a good reason to restrict the event types in the builder

brezinajn:
You're right. That'd work only when specified as argument of the function not within the builder (`@BuilderInference` could help there?)

brezinajn:
When specified as argument it works fine btw.
``` Kotlin
sealed class EventType<T: PostgresAction>{
    object All: EventType<PostgresAction>()
    object Insert: EventType<PostgresAction.Insert>()
    object Update: EventType<PostgresAction.Update>()
    object Delete: EventType<PostgresAction.Delete>()
}

fun <T: PostgresAction>foo(type: EventType<T>): T = TODO()
```

image

jan-tennert:
Ah I know what you mean. I will to do that. Also other question:
For presence you would normally have two "events". In the one you get data from other clients when they "join" the channel. And from the other you get data when they "leave". (Can be used for something like a "I'm online" status thing).
Should I use two flows for both events?

jan-tennert:
> When specified as argument it works fine btw.
> 
> ```kotlin
> sealed class EventType<T: PostgresAction>{
>     object All: EventType<PostgresAction>()
>     object Insert: EventType<PostgresAction.Insert>()
>     object Update: EventType<PostgresAction.Update>()
>     object Delete: EventType<PostgresAction.Delete>()
> }
> 
> fun <T: PostgresAction>foo(type: EventType<T>): T = TODO()
> ```
> 
> image

okay that was only possible if I had a function for changing the event in the builder. Now I could either to that or remove the event property completely and set the event type by parsing the generic parameter.

brezinajn:
> Ah I know what you mean. I will to do that. Also other question: For presence you would normally have two "events". In the one you get data from other clients when they "join" the channel. And from the other you get data when they "leave". (Can be used for something like a "I'm online" status thing). Should I use two flows for both events?

If I understand the question correctly, I'd use a single flow for both types of presence messages.

brezinajn:
> > When specified as argument it works fine btw.
> > ```kotlin
> > sealed class EventType<T: PostgresAction>{
> >     object All: EventType<PostgresAction>()
> >     object Insert: EventType<PostgresAction.Insert>()
> >     object Update: EventType<PostgresAction.Update>()
> >     object Delete: EventType<PostgresAction.Delete>()
> > }
> > 
> > fun <T: PostgresAction>foo(type: EventType<T>): T = TODO()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> 
> okay that was only possible if I had a function for changing the event in the builder. Now I could either to that or remove the event property completely and set the event type by parsing the generic parameter. So either
> 
> ```kotlin
> val flow = channel.postgrestChangeFlow<PostgresAction.Select() { ... }
> ```
> 
> or
> 
> ```kotlin
> val flow = channel.postgrestChangeFlow {
>     setEvent(EventType.Select)
> }
> ```

Specifying via reified type param is actually a nice idea.

jan-tennert:
I updated the pr's description, what do you think of the presence syntax?

brezinajn:
Looks good

PR #13, opened 2022-10-06
Fix incorrect PostgrestRequest.RPC.method

brezinajn:
Method for RPC should be POST.


jan-tennert:
Included in 0.4.0-beta-2

PR #16, opened 2022-10-17
Add ios support

jan-tennert:
This PR adds support for IOS for supabase-kt.

Goals
- [x] Full support for all networking methods (should be working fine via https://ktor.io/docs/http-client-engines.html#darwin)
- [x] Support for OAuth


jan-tennert:
Can't be merged until I have access to the Github Actions

Issue #20, opened 2022-10-31
[Feature request]: provide a BOM dependency


brezinajn:
### General Info

- [X] I installed the latest version of Supabase Kt
- [X] I checked for similar feature requests

### Feature request

It'd be nice to provide a BOM dependency to simplify dependency version management.

### Usecase

It's an easy way to avoid dependency version misconfiguration. You probably never want to use a different version of one module than the others.

brezinajn:
Works great. Thanks

PR #21, opened 2022-10-31
Add support for BOM

jan-tennert:

This PR adds support for a BOM dependency (#20), which you can include in gradle using

```kotlin
implementation(platform("io.github.jan.supabase:bom:VERSION"))
implementation("io.github.jan.supabase:gotrue-kt")
//and other supabase-kt dependencies
```
**or Kotlin MPP projects:**
```kotlin
implementation(project.dependencies.platform("io.github.jan.supabase:bom:VERSION"))
```

This PR should work as is, tested on maven local

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -1,0 +1,103 @@
+plugins {
+    `java-platform`
+}
+
+group = "io.github.jan-tennert.supabase"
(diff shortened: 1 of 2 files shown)


jan-tennert:
@brezinajn Does that look right?

brezinajn:
Looks good to me.

PR #22, opened 2022-11-01
untrack local.properties

brezinajn:

`local.properties` should not be tracked

`local.properties` is being tracked

## What is the new behavior?

`local.properties` is tracked no more


jan-tennert:
You should probably also add **local.properties** to the gitignore just in case 👍

jan-tennert:
Nevermind its already, weird

PR #27, opened 2022-11-16
Improve error handling for all APIs

jan-tennert:

Feature

All apis throw a RestException (mostly) including the response payload

## What is the new behavior?

All main plugins have to implement a new method: **parseErrorResponse** which is called when the library receives a response with the status code **400..499**.
This method returns a **RestException** which is now a sealed class, which currently can be of type:
- UnauthorizedException (for auth-related issues)
- BadRequestException (for things like missing or wrong fields etc.)
- NotFoundException (should be self-explanatory)
- UnknownRestException (for the rest)

The plugins parse the corresponding response payload and status code and return a matching exception which is thrown by the core library

On top of that supabase-kt throws two exceptions when having network problems:
- HttpRequestTimeoutException from Ktor (can be configured in the supabase client builder)
- HttpRequestException for all other network related problems

### DIFF
--- a/Functions/src/commonMain/kotlin/io/github/jan/supabase/functions/EdgeFunction.kt
+++ b/Functions/src/commonMain/kotlin/io/github/jan/supabase/functions/EdgeFunction.kt
@@ -2,8 +2,8 @@ package io.github.jan.supabase.functions
 
 import io.github.jan.supabase.SupabaseClient
 import io.github.jan.supabase.annotiations.SupabaseInternal
+import io.github.jan.supabase.exceptions.RestException
 import io.ktor.http.Headers
-import io.ktor.http.HeadersBuilder
 import io.ktor.http.HttpHeaders
 
 /**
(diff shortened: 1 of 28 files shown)


jan-tennert:
The exceptions currently look like this:
```bash
io.github.jan.supabase.exceptions.UnauthorizedRestException: invalid JWT: unable to parse or verify signature, token is expired by 3698h47m36s: null
URL: https://arnyfaeuskyqfxkvotgj.supabase.co/auth/v1/user
Headers: ...
Http Method: GET
```

brezinajn:
Is there gonna be support for status unrelated exceptions? Timeout, dns resolve issues etc.

jan-tennert:
**For timeouts:** Could be implemented relatively easy via ktor
**For dns resolve issues:** Not sure as the ktor engines throw exceptions depending on the engine/platform. Could add a try/catch for the whole request and throw a custom exception

jan-tennert:
I could also add the possibility to change the request timeout in the supabaseClientBuilder

jan-tennert:
Ktor now throws a HttpRequestTimeoutException on timeout which can be catched easily

jan-tennert:
@brezinajn Should I leave the exceptions for all network related problems as is, because as a library I don't know which exceptions the platform engine is going to throw, or should I catch all exceptions and throw a custom exception like **HttpRequestException** (or something like that)

brezinajn:
That sounds reasonable.

PR #66, opened 2023-03-12
Sign Up & Login with IDToken - GoTrue

ferhatwi:

Feature

Sign Up & Login with IDToken is not possible .

## What is the new behavior?

Sign Up & Login with IDToken is now possible.

## Additional context

There is a Supabase JS method that allows to sign up & login with IDToken, can be shown in the link: https://github.com/supabase/gotrue-js/blob/d10a5b6b010c72dc33709dd1d6c6c34ad762e302/src/GoTrueClient.ts#L387

### DIFF
--- a/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/IDToken.kt
+++ b/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/IDToken.kt
@@ -12,16 +12,15 @@ import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.decodeFromJsonElement
 
 /**
- * Authentication method with email and password
+ * Authentication method with id token, client id, provider and optionally nonce
  */
-object Email : DefaultAuthProvider<Email.Config, Email.Result> {
+object IDToken : DefaultAuthProvider<IDToken.Config, IDToken.Result> {
 
     @Serializable(with = DefaultAuthProvider.Config.Companion::class)
-    data class Config(var email: String = "", var password: String = ""): DefaultAuthProvider.Config()
+    data class Config(var id_token: String = "", var client_id: String = "", var provider: String = "", var nonce: String? = null): DefaultAuthProvider.Config()
     @Serializable
     data class Result(
         val id: String,
-        val email: String,
         @SerialName("confirmation_sent_at") val confirmationSentAt: Instant,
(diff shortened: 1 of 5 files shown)


GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/IDToken.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+
+/**
+ * Authentication method with id token, client id, provider and optionally nonce
+ */
+object IDToken : DefaultAuthProvider<IDToken.Config, IDToken.Result> {
+
+    @Serializable(with = DefaultAuthProvider.Config.Companion::class)
+    data class Config(var id_token: String = "", var client_id: String = "", var provider: String = "", var nonce: String? = null): DefaultAuthProvider.Config()
```
jan-tennert, line 20: Probably best to stick with camel case:

```suggestion
    data class Config(@SerialName("id_token") var idToken: String = "", @SerialName("client_id") var clientId: String = "", var provider: String = "", var nonce: String? = null): DefaultAuthProvider.Config()
```

GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/Phone.kt
```diff
@@ -17,7 +17,7 @@ import kotlinx.serialization.json.decodeFromJsonElement
 object Phone : DefaultAuthProvider<Phone.Config, Phone.Result> {
 
     @Serializable(with = DefaultAuthProvider.Config.Companion::class)
-    data class Config(var phoneNumber: String = ""): DefaultAuthProvider.Config()
+    data class Config(var phoneNumber: String = "", var password: String = "",): DefaultAuthProvider.Config()
```
jan-tennert, line 20:
```suggestion
    data class Config(var phoneNumber: String = "", var password: String = ""): DefaultAuthProvider.Config()
```

jan-tennert:
One thing to consider:
- As only Google and Apple are supported, you could make an sealed interface (IdTokenProvider?) which the OAuthProviders Google and Apple inherit from and set the IDTokenConfig#provider type to IdTokenProvider to have that extra type safety.

PR #80, opened 2023-03-30
Selection helper for Postgrest

ferhatwi:

Feature

Does not exists

## What is the new behavior?

Method:
```kotlin
postgrest["some_table"].select<SomeDataClass>()
```
can now be used to automatically select the columns from constructor parameters of SomeDataClass.

## Additional context

I added this feature to `androidMain`. I don't know if I should have added it to `jvmMain` as well or just to `jvmMain`.

This function could also return `PostgrestSelectResult<SomeDataClass>`, but I think it needed more change of the code structure.

### DIFF
--- a/Postgrest/src/androidMain/kotlin/io/github/jan/supabase/postgrest/query/select.kt
+++ b/Postgrest/src/androidMain/kotlin/io/github/jan/supabase/postgrest/query/select.kt
@@ -1,0 +1,33 @@
+package io.github.jan.supabase.postgrest.query
+
+import io.github.jan.supabase.exceptions.HttpRequestException
+import io.github.jan.supabase.exceptions.RestException
+import io.github.jan.supabase.postgrest.request.PostgrestRequest
+import io.ktor.client.plugins.*
+import kotlin.reflect.full.primaryConstructor
+
+/**
+ * Executes vertical filtering with select on [PostgrestBuilder.table]
+ *
+ * T is the desired data class as a result of the selection.
+ * @param head If true, select will delete the selected data.
+ * @param count Count algorithm to use to count rows in a table.
+ * @param single If true, select will return a single row. Throws an error if the query returns more than one row.
+ * @param filter Additional filtering to apply to the query
+ * @return PostgrestResult which is either an error, an empty JsonArray or the data you requested as an JsonArray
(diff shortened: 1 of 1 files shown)


jan-tennert:
I'm still trying to understand the use case for this. Couldn't we just introduce a list parameter for the select method (rather than a string parameter, because selecting only a few rows is already possible), or what's special about this data class? Is it made only for this query or does it have another purpose?

ferhatwi:
Lets say you have a table called `test`. The table has 3 columns: `id`, `created_at` and `value`. If I only need id and value, I would run this code:
```kotlin
@Serializable
data class Test(val id: String, val value: String)

suspend fun main() {
    val result = client.postgrest["test"].select()

    println(result.decodeSingle<Test>())
}
```
and this would throw error: 
```
Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Encountered an unknown key 'created_at'.
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
```

So I would have to change the main function to either this:
```kotlin
suspend fun main() {
    val result = client.postgrest["test"].select("id,value")

    println(result.decodeSingle<Test>())
}
```
or this:
```kotlin
suspend fun main() {
     val result = client.postgrest["test"].select()

     println(result.decodeSingle<Test>(Json { ignoreUnknownKeys = true }))
}
```
But in the feature that I have provided, running this:
```kotlin
suspend fun main() {
     val result = client.postgrest["test"].select<Test>()

     println(result.decodeSingle<Test>()
}
```
would be enough. 

If we also implement the feature where: 

- `select` method returns `PostgrestSelectResult<Test>` instead of `PostgrestResult`, and 
- `decodeSingle` decodes to `Test` instead of some generic type, 

we don't even need to specify `<Test>`  in the method `result.decodeSingle<Test>()`. 

So this feature is nothing spectacular, it's a helpful tool that eliminates the need for column selection, as using data classes basically means a column selection.

jan-tennert:
For this case it may work and we could think about falling back to the serial names on JS and Native

jan-tennert:
I would propose a refactor for this, supporting the feature you requested, and close this PR, if you don't mind

ferhatwi:
Okay, thanks!

PR #96, opened 2023-04-13
Update docs to reflect the real usage of Supabase URL when provide each service

hieuwu:

Update doc to avoid misunderstanding usage of Supabase URL

Currently, the docs for GoTrue and Postgrest are telling developer to set url when setup service instance as `"https://your.postgrest.url.com"`. However, according to the document in dashboard and it must have the endpoint for each service for example `/auth` or `/rest`

## What is the new behavior?
Update usage of Supabase url to be included endpoint of each service
## Additional context

### DIFF
--- a/GoTrue/README.md
+++ b/GoTrue/README.md
@@ -27,7 +27,7 @@ val client = createSupabaseClient {
 
 or create standalone module
 ```kotlin
-val gotrue = standaloneSupabaseModule(GoTrue, url = "https://your.gotrue.url.com", apiKey = "your-api-key")
+val gotrue = standaloneSupabaseModule(GoTrue, url = "https://your.gotrue.url.com/auth/v1", apiKey = "your-api-key")
 ```
 
 # Usage
--- a/Postgrest/README.md
+++ b/Postgrest/README.md
@@ -27,7 +27,7 @@ val client = createSupabaseClient {
 
 or create standalone module
 ```kotlin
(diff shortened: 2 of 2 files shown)


jan-tennert:
In any case, I agree this documentation should be improved but if you add auth/v1, you should also change the url itself to look like a supabase url (or change it completely)

hieuwu:
> The original idea for this was when you have some custom postgrest instance, you could easily create the specific module without a supabase url etc. (Thats why the url even looks like this and not like 'https://id.supabase.co')
> For supabase projects it is recommended to use the supabase client directly as you only need to specify your supabase url one time, and you have a lot more freedom when using more modules later. On top of that, you need GoTrue for authenticated requests.

Yeah you are right. I’m just not aware of the custom Supabase. I think we can just leave it as is for now and will try to provide better document later for the whole use case. And just notice that we should provide the whole client instead of each standalone module 😄 Thank you for this.

jan-tennert:
> > The original idea for this was when you have some custom postgrest instance, you could easily create the specific module without a supabase url etc. (Thats why the url even looks like this and not like 'https://id.supabase.co')
> > For supabase projects it is recommended to use the supabase client directly as you only need to specify your supabase url one time, and you have a lot more freedom when using more modules later. On top of that, you need GoTrue for authenticated requests.
> 
> Yeah you are right. I’m just not aware of the custom Supabase. I think we can just leave it as is for now and will try to provide better document later for the whole use case. And just notice that we should provide the whole client instead of each standalone module 😄 Thank you for this.

No problem, I agree my documentation can be confusing

PR #205, opened 2023-06-21
Add support for custom serialization

### LINKED ISSUE #203: [Feature request]: Add support for custom serialization
### General Info

- [X] I installed the latest version of Supabase Kt
- [X] I checked for similar feature requests

### Feature request

There is currently no support for custom serializers in supabase-kt. You can use `PostgrestResult#body.toString()` or `PostgresAction#record.toString()` and decode it yourself, but that comes with a lot of boilerplate.

See #202 

### Usecase

Ability to adapt to your current serialization method

jan-tennert:

Feature (closes #203)

There is no convenient way of integrating other serialization frameworks into supabase-kt. You can use get the raw body string for some responses (like Postgrest & Realtime) but for all other modules, it is not possible.

## What is the new behavior?

Some modules will have a new configuration property called `serializer`. This property accepts an implementation of the `SupabaseSerializer` interface.

Here is an example using Moshi:
```kotlin
data class User(val name: String, val age: Int)

class MoshiSerializer(private val moshi: Moshi): SupabaseSerializer {

    @OptIn(ExperimentalStdlibApi::class)
    override fun <T : Any> encode(type: KType, value: T): String = moshi.adapter<T>(type).toJson(value)

    @OptIn(ExperimentalStdlibApi::class)
    override fun <T : Any> decode(type: KType, value: String): T {
        val adapter = moshi.adapter<T>(type)
        return adapter.fromJson(value) ?: error("Failed to decode $value")
    }


}

suspend fun main() {
    val moshiSerializer = MoshiSerializer(
        Moshi.Builder()
            .addLast(KotlinJsonAdapterFactory())
            .build()
    )
    val client = createSupabaseClient(
        supabaseUrl = "",
        supabaseKey = ""
    ) {
        install(Postgrest) {
            serializer = moshiSerializer
        }
    }
    client.postgrest["users"].insert(User("Jan", 20))
}
```

## Additional context

Suggestions are welcome!


yoobi:
Hello ! Thanks for the development, I've picked up your `MoshiSerializer` but I'm getting `Required: Class<TypeVariable(T)!>!, Found:KType` in the snippet below
```
class MoshiSerializer(private val moshi: Moshi): SupabaseSerializer {

    @OptIn(ExperimentalStdlibApi::class)
    override fun <T : Any> encode(type: KType, value: T): String = moshi.adapter<T>(type).toJson(value)
                                                                                                                                               ^^^
    @OptIn(ExperimentalStdlibApi::class)
    override fun <T : Any> decode(type: KType, value: String): T {
        val adapter = moshi.adapter<T>(type)
                                                                ^^^
        return adapter.fromJson(value) ?: error("Failed to decode $value")
    }
}
```
Is there something I've done wrong ?

jan-tennert:
yea thought about that, might do that for the full 1.1.0 release, but good to know that it works!

yoobi:
Thanks again for considering my request :)

jan-tennert:
> 

Also btw, you don't need this block:

```kotlin
httpEngine = HttpClient(OkHttp) {
            install(WebSockets)
        }.engine
```
Ktor will search for a http engine automatically and use it (if there is one in your classpath). Realtime will also install the WebSockt functionality, so that is also not needed.

yoobi:
Oh you're right ! Thanks for pointing that out

PR #277, opened 2023-09-03
Separate PostgrestRequest, move Rpc Execution to the same place with other requests and cover unit test

hieuwu:
### Analyst on PostgrestRequest

Currently, this interface contain set of properties of a request, any class that extends this interface will share `execute` 
 and `asPostgrestResult` methods with the same implementation.
For a request class, it should only contain the information of a request and the execution and parsing result responsibility should belong to other class. A request should not execute and contain other logic related to  This will help separate the responsibilities between classes.
rpc is considered as a kind of request to execute. However, it is implemented as an extension function of Postgrest. It should be somehow treated as other rest requests

#### Things that we can improve:
- `PostgrestRequest` hold many responsibilities, we can break it in to smaller classes and files
- rpc is implemented as an extension function of `Postgrest`, it should be treated as other kinds of requests
- `PostgrestResult` class should be in `result` package and not query

#### Solution
- Create `RequestExecutor`, its responsibility is receiving a request, executing and parsing the result
- Create set of request classes to hold their properties
- Move rpc to the same place with other kinds of requests
- Restructure PostgrestResult to be in `result` package
- Cover unit tests for all request classes

#### Test coverage
Screenshot 2023-09-14 at 22 48 35


jan-tennert:
Thank you for this PR! A few things:
- The reason why `rpc` was an extension function of `Postgrest` is
    1. Because you don't specify a table when calling a DB Function. The syntax with this PR would look like this: 
     ```kotlin
     client.postgrest[""].rpc("my_function")
     ```
     2. Postgrest is an interface, which means there can be no inline function (which is required for one of the `rpc` functions)
- Detekt probably mostly fails because of missing documentation. This means that every public method/class/property has to be documented. I would suggest that you mark most of the new classes/interfaces as internal as they aren't used outside the module anyway. (This will disable the documentation check). You can also run `gradlew detektAll` yourself, to see all errors.

hieuwu:
> Thank you for this PR! A few things:
> 
> * The reason why `rpc` was an extension function of `Postgrest` is
>   
>   1. Because you don't specify a table when calling a DB Function. The syntax with this PR would look like this:
>   
>   ```kotlin
>   client.postgrest[""].rpc("my_function")
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   
>   2. Postgrest is an interface, which means there can be no inline function (which is required for one of the `rpc` functions)
> * Detekt probably mostly fails because of missing documentation. This means that every public method/class/property has to be documented. I would suggest that you mark most of the new classes/interfaces as internal as they aren't used outside the module anyway. (This will disable the documentation check). You can also run `gradlew detektAll` yourself, to see all errors.

Thanks Jan. Didn’t know the actual syntax of rpc in the change. It looks like the old implementation is good for the use case. Will try some other ways to see more options.
Do you know which is preferred library for mocking dependencies when writing test for KMP library Jan?

jan-tennert:
There is [mockative](https://github.com/mockative/mockative) or [MocKMP](https://github.com/kosi-libs/MocKMP) but I have never used any of them before. There is also a Kotlin Multiplatform Testing library with more advanced features: [kotest](https://kotest.io/), but I haven't used that either.

hieuwu:
> There is [mockative](https://github.com/mockative/mockative) or [MocKMP](https://github.com/kosi-libs/MocKMP) but I have never used any of them before. There is also a Kotlin Multiplatform Testing library with more advanced features: [kotest](https://kotest.io/), but I haven't used that either.

Thank you Jan. Actually I've just tried mockative but it causes some generate error. Will play around with it and kotest

hieuwu:
It seems pretty fine with the separation for now please take a look @jan-tennert . I tried mockative and mockkmp but still faced some issues regarding mocking dependencies and build issue, will try more in another PR.

Postgrest/src/commonMain/kotlin/io/github/jan/supabase/postgrest/Postgrest.kt
```diff
@@ -45,6 +32,12 @@ import kotlinx.serialization.json.JsonElement
  */
 sealed interface Postgrest : MainPlugin<Postgrest.Config>, CustomSerializationPlugin {
 
+    /**
+     * Instance of Authenticated Supabace Api to interact with Supabase APIs
+     * @see io.github.jan.supabase.network.SupabaseApi
+     */
+    val api: AuthenticatedSupabaseApi
```
jan-tennert, line 39: Not sure if the api object should be exposed, as no other plugin does it

jan-tennert (approved):
Looks good! One small thing.

Postgrest/src/commonMain/kotlin/io/github/jan/supabase/postgrest/Postgrest.kt
hieuwu: Sure we don't need to expose api object. Thanks for pointing this out Jan

PR #284, opened 2023-09-10
Update .gitignore to avoid generated code

hieuwu:
I just pulled latest code in my current repository and it generate thousands of files so I add them to git ignore

Improve gitignore

I face there are too many generated fiels
Screenshot 2023-09-10 at 16 49 17
Screenshot 2023-09-10 at 16 48 42
Screenshot 2023-09-10 at 16 44 43

## What is the new behavior?
Generated change should be avoided

## Additional context
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,7 @@
 /plugins/ApolloGraphQL/build
 /plugins/ComposeAuthUI/build
 /plugins/ComposeAuth/build
+/serializers/Jackson/build
+/serializers/Moshi/build
 .idea
 local.properties
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@
 /GoTrue/build/
 /Functions/build/
 /plugins/ApolloGraphQL/build
-/plugins/ApolloGraphQL/build/
 /plugins/ComposeAuthUI/build/
 /plugins/ComposeAuth/build/
 /serializers/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,5 +11,8 @@
 /GoTrue/build/
 /Functions/build/
 /plugins/ApolloGraphQL/build
+/plugins/ComposeAuthUI/build/
+/plugins/ComposeAuth/build/
+/serializers/
```
jan-tennert, line 16: Now the whole serializers directory gets ignored, probably have to write out the paths too

jan-tennert (changes requested):

.gitignore
hieuwu: > Now the whole serializers directory gets ignored, probably have to write out the paths too

Just updated. What do you mean by mentioning _write out the paths_ ?

jan-tennert: No, I just mean the serializers directory (/serializiers/Jackson/build/ etc.)

hieuwu: > No, I just mean the serializers directory (/serializiers/Jackson/build/ etc.)

Updated. Just wonder why do we need to specify for each serializer? If we introduce more, we need to update ignore again?

jan-tennert: Probably, the serializers also generate build files

jan-tennert (approved):

.gitignore
hieuwu: Make sense

PR #329, opened 2023-11-05
Fix documentation of the BucketApi in the Storage module

shubhamsinghshubham777:

This PR fixes the following issues in the documentation of the BucketApi in the Storage module:
1. Fix incorrect variable names
2. Add missing import

The documentation mentions variables that do not exist anymore (e.g. in the `createBucket` method, the documentation mentions a param called `name` that has been replaced with `builder` now)

Screenshot 2023-11-05 at 1 29 04 PM

## What is the new behavior?

The updated documentation fixes all issues mentioned above, for example:

Screenshot 2023-11-05 at 1 47 16 PM


jan-tennert (approved):
Thanks!

Issue #342, opened 2023-11-13
Deprecate `Returning.MINIMAL` and migrate to a similar syntax to JS library


dshukertjr:
Currently, inserting and retrieving records looks something like this (from the docs): 

```kotlin
val city = City(name = "The Shire", countryId = 554)
supabase.postgrest["cities"].insert(city, returning = Returning.MINIMAL)
```

It would be better if we could have a syntax that is more similar to the JS client. 

```ts
const { data, error } = await supabase
  .from('countries')
  .insert({ id: 1, name: 'Denmark' })
  .select()
```

JS docs: https://supabase.com/docs/reference/javascript/insert?example=create-a-record-and-return-it

jan-tennert:
Just a small note:
inserting and retrieving would be something like this:
```kotlin
val city = supabase.postgrest["cities"].insert(city).decodeSingle<City>()
```
Using `returning = Returning.MINIMAL` would be the same as not using `select()` in JS

dshukertjr:
Ah, right. Mine was the opposite 😂

jan-tennert:
Drafted the new syntax here: #353

PR #353, opened 2023-11-20
Refactor Postgrest syntax and implementation

### LINKED ISSUE #352: Rename `foreignTable` to `referencedTable` in Postgrest
# Feature request

## Is your feature request related to a problem? Please describe.

`foreignTable` is now renamed to `referencedTable` in the JS SDK. It would be nice to align symbol name in Kotlin SDK before the upcoming version update as well. 

Here is the JS SDK implementation: https://github.com/supabase/postgrest-js/pull/493#issuecomment-1817313103

The JS implementation is deprecating the existing `foreignTable` parameter and adding `referencedTable` so that there are no breaking changes, but since we are doing a major version upgrade, it should be fine to just rename `foreignTable` to `referencedTable` without deprecation.

jan-tennert:

Feature (closes #352 and #342) 

## What is the old behavior?

The old syntax looks something like this:
```kotlin
val city = City(name = "The Shire", countryId = 554)
supabase.postgrest["cities"].insert(city, returning = Returning.MINIMAL)
```
```kotlin
supabase.postgrest["countries"].update(newCity, returning = Returning.REPRESENTATION, count = Count.EXACT) {
   eq("id", 1)
}
```

## What is the new behavior?

This PR refactors the Postgrest syntax. All parameters in the query methods (insert, update, delete etc.) have been moved to the new request builder. Here is an example (this is just to show all methods, you'll most likely not use all of them together):
```kotlin
val result = supabase.postgrest["message"].update(newMessage) { //you can also use supabase.from("message")
    select()
    count(Count.ESTIMATED)
    single()
    order("user_id", Order.ASCENDING, referencedTable = "users")
    csv()
    geojson()
    explain(analyze = true)
    filter {
        eq("id", 1)
    }
}
println(result.data)
println(result.decodeSingle<Message>())
println(result.countOrNull())
println(result.headers)
```

## Additional context


dshukertjr:
Love the new syntax! Thanks for this update!

Issue #290, opened 2023-09-14
[Feature request]: Add iOS to demos


jan-tennert:
### General Info

- [X] I installed the latest version of Supabase Kt
- [X] I checked for similar feature requests

### Feature request

It would be good if all demos which have android as a target (excluding the android login demo) also target iOS. I don't think that would be too hard as the common modules don't use JVM-related stuff as most of the demos also target JS. I can't do that myself as I don't have a Mac to test it. Help appreciated.

### Usecase

*-*

hieuwu:
Is this issue just to try and run the sample project in iOS emulator to make sure it works Jan?

jan-tennert:
> Is this issue just to try and run the sample project in iOS emulator to make sure it works Jan?

Ideally the `chat-demo`, `file-upload`, `multiplatform-deeplinks`, and `multi-factor-authentication`  demos get a (target) module for iOS. Which should just be adding some expect/actual methods in common code and in the iOS module is an iOS app just using compose like android.

hieuwu:
Thanks for the context. I'll give it a try

PR #397, opened 2023-12-25
Set up iOS for chat demo

hieuwu:
Set up iOS projet for chat demo app
To resolve https://github.com/supabase-community/supabase-kt/issues/290
iOS version is not available

## What is the new behavior?
Chat Demo iOS version is available
Screenshot 2023-12-27 at 15 04 39

## Additional context
How to set up with XCode 15 and run the app on iOS 17
Follow this tutorial
[Make your cros-platform work on iOS](
https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-integrate-in-existing-app.html#make-your-cross-platform-application-work-on-ios)

However, for **Framwork Search Paths**, the value should be `$(SRCROOT)/../common/build/xcode-frameworks/$(CONFIGURATION)/$(SDK_NAME)` , **Other linker flags** is `"$(inherited) -framework common"`

**Note**:
- First build the whole project in Android Studio
- If there's no iOS build config available in Android, open XCode then press build button


demos/chat-demo-mpp/gradle/libs.versions.toml
```diff
@@ -1,16 +1,16 @@
...
-ktor = "2.3.1"
+koin = "3.4.3"
+ktor = "2.3.7"
 
 androidx-core = "1.10.1"
 androidx-compat = "1.6.1"
 androidx-lifecycle = "2.6.1"
 androidx-activity = "1.7.2"
-
+org-jetbrains-kotlin-jvm = "1.9.10"
```
jan-tennert, line 13: Two things:
- We can use the newest possible versions if we are updating them anyways (Kotlin 1.9.21, Compose 1.5.11)
- Is there a reason kotlin-jvm has a separate version entry?

jan-tennert (approved):
Other than that, looks good thanks! We might need to do some cleaning up in the common code (because of Compose and supabase-kt changes), but that is for another PR.

demos/chat-demo-mpp/gradle/libs.versions.toml
hieuwu: - Let me try the latest posible versions
- No it's not. It should be the same version with Kotlin. I'll update it

hieuwu: Just updated. All good now

hieuwu:
> Other than that, looks good thanks! We might need to do some cleaning up in the common code (because of Compose and supabase-kt changes), but that is for another PR.

Yeah let me make the demos work then go back to update the change accordingly

jan-tennert (approved):

PR #412, opened 2024-01-03
Fix crash when parsing params after hash in URL

### LINKED ISSUE #411: [Bug]: JS hash url parsing fails in GoTrue
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/wiki/Troubleshooting) page for similar problems
- [X] I enabled [logging](https://github.com/supabase-community/supabase-kt/wiki/Logging) and checked the logs

### Version(s)

2.0.3

### Target(s)

JS (IR)

### What happened? (include your code)

Application crashes with an uncaught exception and without any reason.
Seems like this [file](https://github.com/supabase-community/supabase-kt/blob/master/GoTrue/src/jsMain/kotlin/io/github/jan/supabase/gotrue/setupPlatform.kt) is responsible for the issue. The function `checkForHash` is called every time a new url is loaded and expects the url to contain "&" and "=" characters but it doesn't.

### Steps To Reproduce (optional)

1. Initialize Supabase Client
2. Navigate to a page using hash based routing (www.mysite.com/#/users)
3. Navigate to another page using hash based routing (www.mysite.com/#/doctors)
4. Go back 

```shell
index: 1, size: 1
IndexOutOfBoundsException: index: 1, size: 1
    at protoOf.checkElementIndex_r3t1bm_k$ (webpack-internal:///./kotlin/kotlin-kotlin-stdlib.js:14291:13)
    at rangeCheck (webpack-internal:///./kotlin/kotlin-kotlin-stdlib.js:8736:31)
    at protoOf.get_c1px32_k$ (webpack-internal:///./kotlin/kotlin-kotlin-stdlib.js:8765:28)
    at setupPlatform$checkForHash (webpack-internal:///./kotlin/supabase-kt-gotrue-kt.js:12209:51)
    at eval (webpack-internal:///./kotlin/supabase-kt-gotrue-kt.js:12273:7)
```

dancing-koala:

Bug fix ( closes #411 )

Application crashes with an IndexOutOfBoundsException when navigating to a URL containing a hash but not parameters.
Example from the issue : http://www.mysite.com/#/users

## What is the new behavior?

The application doesn't crash when the value after the hash does not contain any param.

## Additional context

This will happen for sure when using hash based navigation.

### DIFF
--- a/GoTrue/src/jsMain/kotlin/io/github/jan/supabase/gotrue/setupPlatform.kt
+++ b/GoTrue/src/jsMain/kotlin/io/github/jan/supabase/gotrue/setupPlatform.kt
@@ -14,7 +14,15 @@ actual fun Auth.setupPlatform() {
 
     fun checkForHash() {
         if(window.location.hash.isBlank()) return
-        val map = window.location.hash.substring(1).split("&").associate {
+
+        val afterHash = window.location.hash.substring(1)
+
+        if(!afterHash.contains('=')) {
+            // No params after hash, no need to continue
+            return
+        }
+
+        val map = afterHash.split("&").associate {
             it.split("=").let { pair ->
                 pair[0] to pair[1]
             }


jan-tennert (approved):
Thank you!

PR #421, opened 2024-01-08
User filter in realtime list data flow initial value

jollygreenegiant:

Fixes a bug where the filter was not being used in the new realtime list data flow

`postgresListDataFlow()` accepts a filter, but doesn't use it when retrieving the initial value, so it just gets the entire table.

## What is the new behavior?

`postgresListDataFlow()` now applies the filter on the initial data retrieval


jan-tennert (approved):
Thank you! I changed the base branch, as the 2.1.0 branch is currently used for the next version instead of development!

PR #430, opened 2024-01-16
Remove error for Realtime when a channel is already connected

jollygreenegiant:

Update Realtime channels to not throw an error when trying to connect to a channel that is already connected.

Currently trying to connect with a channel that is already connected will throw an `IllegalStateException`

## What is the new behavior?

Now, the `connect()` function will just return if the channel is already connected.  

## Additional context

The reasoning behind this is that if you are trying to connect to a channel, and that channel is already connected, then you're in the state you want to be in and there's no need to throw an error.


jan-tennert (approved):
Thanks!

PR #431, opened 2024-01-19
Add GoTrue lifecycle observers on main thread

jollygreenegiant:

Fixes a crash where instrumented tests would crash when building the Auth library.

I was getting this crash when running instrumented tests on my app:

```
Caused by java.lang.IllegalStateException
    Method addObserver must be called on the main thread
    androidx.lifecycle.LifecycleRegistry.enforceMainThreadIfNeeded (LifecycleRegistry.kt:304)
    androidx.lifecycle.LifecycleRegistry.addObserver (LifecycleRegistry.kt:181)
    io.github.jan.supabase.gotrue.SetupPlatformKt.addLifecycleCallbacks (setupPlatform.kt:33)
    io.github.jan.supabase.gotrue.SetupPlatformKt.setupPlatform (setupPlatform.kt:25)
    io.github.jan.supabase.gotrue.AuthImpl.init (AuthImpl.kt:80)
    io.github.jan.supabase.SupabaseClientImpl.<init> (SupabaseClient.kt:96)
    io.github.jan.supabase.SupabaseClientBuilder.build (SupabaseClientBuilder.kt:79)
```

Android requires that `Lifecycle.addObserver()` be called on the main thread, so this was causing crashes on instrumented tests (but not on real devices that I could find) when building a SupabaseClient that uses the Auth library

## What is the new behavior?

`addObserver()` is now called on the main thread, which should eliminate the crash.

## Additional context

https://developer.android.com/reference/androidx/lifecycle/Lifecycle#addObserver(androidx.lifecycle.LifecycleObserver)

### DIFF
--- a/GoTrue/src/androidMain/kotlin/io/github/jan/supabase/gotrue/setupPlatform.kt
+++ b/GoTrue/src/androidMain/kotlin/io/github/jan/supabase/gotrue/setupPlatform.kt
@@ -7,6 +7,7 @@ import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.startup.Initializer
 import co.touchlab.kermit.Logger
 import io.github.jan.supabase.annotations.SupabaseInternal
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 
 private var appContext: Context? = null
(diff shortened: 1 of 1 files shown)


jan-tennert (approved):
Thank you!

Issue #441, opened 2024-01-30
[Bug]: Getting data is not working on safari 16 or lower


MohamedRejeb:
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/wiki/Troubleshooting) page for similar problems
- [X] I enabled [logging](https://github.com/supabase-community/supabase-kt/wiki/Logging) and checked the logs

### Version(s)

2.0.4

### Target(s)

Web

### What happened? (include your code)

Getting data is failing and I'm getting this error in the logs:
```
SyntaxError: Invalid regular expression: invalid group specifier name
RegExp@[native code]
Regex@
Regex_init_$Init$_0@
_init_properties_Utils_kt__jo07cx@
mapToFirstValue@
@
@
@
@
@
@
@
@
promiseReactionJob@[native code]
```

When I remove the `select` method call the error is no longer printing and also I see that this `mapToFirstValue` function is used inside the `select` method.

The error is happening only on safari 16.x or lower I guess (tested on safari 16.1 and 16.2)
It's not happening on safari 17.x

jan-tennert:
I think it might be this regex:
https://github.com/supabase-community/supabase-kt/blob/705dba27cc88bd591bc7f02fa49b9403648a513a/Postgrest/src/commonMain/kotlin/io/github/jan/supabase/postgrest/Utils.kt#L10-L18
Which is used to convert property names to column names in your filter when using the property syntax: `User::createdAt eq value` -> `eq("created_at", value)`.
Can you try to change the property `propertyConversionMethod` in the Postgrest config to `PropertyConversionMethod.NONE`, to check if it is this regex? (By this default the regex approach is used)

MohamedRejeb:
I changed the property `propertyConversionMethod` in the Postgrest config to `PropertyConversionMethod.NONE` but I'm still getting the same error.

MohamedRejeb:
I tied the above function `camelToSnakeCase` in Safari 16.X and it's causing the issue.

PR #445, opened 2024-01-30
Fix regex error in Safari 16.x or lower

MohamedRejeb:

Fixes: #441 

`SNAKE_CASE_REGEX` is not working for Safari 16.x and lower because `(?<=` wasn't supported in Safari.

## What is the new behavior?

I changed the regex and now `SNAKE_CASE_REGEX` in working fine for Safari and other browsers.

## Additional context

I made a small benchmarking for both implementations with long strings to make sure that this change will not cause any issues, and the results are mostly the same.


jan-tennert (approved):
Thanks! I'll include this in the next release.

PR #483, opened 2024-02-19
Fix concurrent calls on Realtime.connect()

### LINKED ISSUE #478: [Bug]: Concurrent connect() to Realtime can lead to instabilities
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/wiki/Troubleshooting) page for similar problems
- [X] I enabled [logging](https://github.com/supabase-community/supabase-kt/wiki/Logging) and checked the logs

### Version(s)

2.1.4 and more

### Target(s)

Every target

### What happened? (include your code)

In my app, I'm subscribing to 2 realtime channels at the same time. With the default config, Realtime `connectOnSubscribe` is set to true.

I realize that my code is working but act really strangely: 
- sometimes, one of my channel stop receiving new events (but no suspect log)
- really often, strange timeouts but my connection is really stable
- sometimes, some crash happens from nowhere
- reading logs, the heart beat seems to be received twice at the same time

I tried to remove one of my channel and just subscribe one and everything is working perfectly fine, no strange behavior.

By looking at the library code, I realize that the two methods `channel.subscribe()` are calling `realtime.connect()` at the same time.

I decided to disable the `connectOnSubscribe` in the config, manually call `realtime.connect()` once, and tweak my code to wait the connection success before subscribing to my channels, and it perfectly works.

Conclusion: we need to prevent concurrent calls of `realtime.connect()`. I'm working on a quick PR rn, I'll submit it today.

### Steps To Reproduce (optional)

_No response_

_No response_

vinceglb:

By introducing a mutex in the Realtime class, we ensure that the `connect()` method is never executed concurrently.

Bug fix #478. If we call multiple times the `connect()` method at the same time, realtime feature can lead to an unpredictable behavior.

## What is the new behavior?

We ensure that the `connect()` method is never executed concurrently.

## Additional context

I used this documentation to help me choose Mutex as a solution: https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion

The method `newSingleThreadContext` is not available on the common target, so `Mutex` seems to be the best option.


jan-tennert (approved):
Thank you!

PR #508, opened 2024-03-09
Add identities parameter

yuchan2215:

This PR adds a parameter to Email.Result.

Currently, Supabase does not have a solution for checking if an account is already registered at the time of new registration, although there is an alternative in JavaScript. However, this cannot be used in Kotlin.
https://github.com/supabase/gotrue-js/issues/513
https://github.com/supabase/supabase-js/issues/296

## What is the new behavior?

By using a parameter, it will be possible to check for duplicates at the time of account registration.

## Additional context

I am Japanese and not proficient in English. I am using ChatGPT and DeepL to translate.

### DIFF
--- a/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/Email.kt
+++ b/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/Email.kt
@@ -1,6 +1,7 @@
 package io.github.jan.supabase.gotrue.providers.builtin
 
 import io.github.jan.supabase.exceptions.SupabaseEncodingException
+import io.github.jan.supabase.gotrue.user.Identity
 import io.github.jan.supabase.supabaseJson
 import kotlinx.datetime.Instant
 import kotlinx.serialization.ExperimentalSerializationApi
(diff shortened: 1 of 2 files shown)


yuchan2215:
It might also be possible to add this to Phone, but I haven't added it because I don't have a contract with an SMS provider and can't test it. If you have the environment for it, it might be good to add it.

GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/providers/builtin/Email.kt
```diff
@@ -39,6 +41,7 @@ data object Email : DefaultAuthProvider<Email.Config, Email.Result> {
     data class Result(
         val id: String,
         val email: String,
+        val identities: List<Identity>?,
```
jan-tennert, line 44:
```suggestion
        val identities: List<Identity>? = null,
```
Just in case, if for some reason no identities entry is there at all.

jan-tennert (commented):
Thanks for the PR! You can add the same thing for `Phone.Result`, it should work exactly the same. Other than that, lgtm!

jan-tennert:
Closed in favor of #528. Thanks anyways!

Issue #509, opened 2024-03-10
[Question]: Should missing Json properties be ignored by default?


hieuwu:
### General info

- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/wiki/Troubleshooting) page for similar problems

### What is your question?

I got the exception of missing Unknown Json property while I already specs which property I want to get from the response.
```kotlin
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 228: Encountered an unknown key 'created_at' at path: $[0].address
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: .....4f391f1aab08","address":null,"created_at":null}]
```

I fixed it by providing the defaultserializer the Json property with needed configuration for KotlinSerializer
```kotlin
    @Provides
    @Singleton
    fun provideSupabaseClient(): SupabaseClient {
        return createSupabaseClient(
            supabaseUrl = BuildConfig.SUPABASE_URL,
            supabaseKey = BuildConfig.API_KEY,
        ) {
            defaultSerializer = KotlinXSerializer(Json { ignoreUnknownKeys = true })
            install(Postgrest)
            install(Auth)
        }
    }
```

Should this configuration be default? As based on each feature, multiple DTO object would be created to query only needed properties.

jan-tennert:
I think it's fine if we ignore unknown keys by default, I don't think there is any downside for this

PR #510, opened 2024-03-10
Ignore exception thrown by missing Json property by default

hieuwu:
Ignore missing json key by default to avoid `kotlinx.serialization.json.internal.JsonDecodingException ` thrown when there's no Json property

- `kotlinx.serialization.json.internal.JsonDecodingException `is thrown when some Json properties are not specified

## What is the new behavior?
- There's no exception thrown when a Json property is missing
https://github.com/supabase-community/supabase-kt/issues/509


jan-tennert (approved):
Thanks!

Issue #543, opened 2024-04-09
[Bug]: Column.Raw not removing line breaks


iruizmar:
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.MD) page for similar problems

### Version(s)

2.2.3

### Kotlin Target(s) and their respective versions

Android 14, at least

### What happened? (include your code)

According to the docs, when querying a table in postgrest, `Columns.Raw` should be able to receive a list of formatted columns with line breaks and it will remove them automatically to make the correct request:
```kt
 Columns.raw("""
    id,
    name,
    cities (
      id,
      name
    )
""".trimIndent()
```

But the compiler is complaining about this:

```(unexpected "\n" expecting "...", field name (* or [a..z0..9_$]), "*" or "count()")```

so I have to remove them manually to make it work:

```kt
.replace("\n", "")
```

### Steps To Reproduce (optional)

1. Try to use `Columns.Raw` when selecting any query with line breaks

PR #544, opened 2024-04-09
Auto remove line breaks on Columns.Raw

iruizmar:

Bug fix

Solves #543 

## What is the new behavior?

`Columns.Raw` now automatically gets rid of line breaks

## Additional context

Let me know if that's not the correct place to introduce it.
Should we also auto add `trimIndent()`?

### DIFF
--- a/Postgrest/src/commonMain/kotlin/io/github/jan/supabase/postgrest/query/Columns.kt
+++ b/Postgrest/src/commonMain/kotlin/io/github/jan/supabase/postgrest/query/Columns.kt
@@ -21,7 +21,7 @@ value class Columns @PublishedApi internal constructor(val value: String) {
          * Select all columns given in the [value] parameter
          * @param value The columns to select, separated by a comma
          */
-        fun raw(value: String) = Columns(value)
+        fun raw(value: String) = Columns(value.replace("\n", ""))
 
         /**
          * Select all columns given in the [columns] parameter


jan-tennert:
Hey, thanks for this PR! After looking at the other client libs, it seems like what they're doing something different:
https://github.com/supabase/postgrest-js/blob/7c3dd21fb962c35e88b88cc7a789fcfcb692af42/src/PostgrestQueryBuilder.ts#L72-L84
This removes newlines but also any whitespaces while allowing whitespaces when quoted.
We should probably also do it like this, maybe define a extension function at the bottom of `Columns.kt` and clean the parameter when creating the colum instance:
```kotlin
private fun String.clean(): String {
    var quoted = false
    val regex = Regex("\\s")
    return this.map {
        if (it == '"') {
            quoted = !quoted
        }
        if (regex.matches(it.toString()) && !quoted) {
            ""
        } else {
            it
        }
    }.joinToString("")
}
```

grdsdev:
Hi all, 

@jan-tennert suggestion is the way to go, to keep same behavior across client libraries.

iruizmar:
That's cool, thanks for checking other client libraries to see their behaviour.
I changed that.
Although, I'm thinking that this is a hidden feature since it's not added to the docs. I can't see it on the JS client library either.

grdsdev:
@iruizmar what do you mean by hidden feature? I see it more as a convenience for using multiline `select` queries.

You wish we had some doc entry mentioning that it is possible to use multiline queries?

grdsdev (approved):

iruizmar:
> @iruizmar what do you mean by hidden feature? I see it more as a convenience for using multiline `select` queries.
> 
> You wish we had some doc entry mentioning that it is possible to use multiline queries? 

We're now allowing spaces only if they are quoted. I think users won't know this feature if it's not documented. Even if I, honestly, can't see a usecase in which you'd want to have a white space.

So yes, adding a section in the docs about this feels the correct approach.

PR #571, opened 2024-04-27
Add .kotlin to .gitignore

MohamedRejeb:
With K2 we have a new `.kotlin` directory generated and needs to be added to `.gitignore`.

https://kotlinlang.org/docs/whatsnew-eap.html#new-directory-for-kotlin-data-in-gradle-projects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /test/
 .idea
 local.properties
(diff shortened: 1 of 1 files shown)


jan-tennert (approved):
Not sure why the tests are failing right now.

Issue #567, opened 2024-04-23
[Bug]: Auth refresh sessions is not working correctly


MohamedRejeb:
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.MD) page for similar problems

### Version(s)

2.3.1

### Kotlin Target(s) and their respective versions

JVM 17

### What happened? (include your code)

Hi,
Refresh sessions is not working as expected when I have two versions of the same app installed with different flavors.
So let's say I have a release version installed with a link of the remote supabase and I'm logged in there. If I install another version of the app with a different flavor, debug version for example, which is linked to my local supabase.
The first app works fine but the second app gets the saved session of the first app and I'm considered as authenticated which is not the case, and all api calls are failing since the token is for the remote supabase not the local one.
```
Info: (Supabase-Auth) Trying to load latest session from storage.
Info: (Supabase-Auth) Successfully loaded session from storage!
```

Either the library should somehow support having different versions of the same app authenticated to different accounts at the same time, or at least when I receive an Unauthenticated exception, it should clear the session and log me out.
Let me know what you think. I can help on this.

Thanks!

jan-tennert:
I guess you could just provide a custom session manager, which saves sessions to separate keys depending on the build mode:
(this is a straight copy from the default one, see text below)
```kotlin
class CustomSessionManager @OptIn(SupabaseInternal::class) constructor(
    settings: Settings = createDefaultSettings()
) : SessionManager {

    @OptIn(ExperimentalSettingsApi::class)
    private val suspendSettings = settings.toSuspendSettings()

    @OptIn(ExperimentalSettingsApi::class, SupabaseInternal::class)
    override suspend fun saveSession(session: UserSession) {
        suspendSettings.putString(SETTINGS_KEY, supabaseJson.encodeToString(session))
    }

    @OptIn(ExperimentalSettingsApi::class, SupabaseInternal::class)
    override suspend fun loadSession(): UserSession? {
        val session = suspendSettings.getStringOrNull(SETTINGS_KEY) ?: return null
        return try {
            supabaseJson.decodeFromString(session)
        } catch(e: Exception) {
            Auth.logger.e(e) { "Failed to load session" }
            null
        }
    }

    @OptIn(ExperimentalSettingsApi::class)
    override suspend fun deleteSession() {
[...]

PR #572, opened 2024-04-27
Add Custom key for SettingsSessionManager and based on supabaseUrl

MohamedRejeb:
Fixes: #567 

I prefer using the full URL instead of the project reference. Because this will work for self-hosted projects as well.


jan-tennert (approved):
Thanks!

grdsdev (approved):

Issue #576, opened 2024-04-30
[Bug]: [Storage] Empty file is created on Dashboard when pass empty `ByteArray` to upload function


hieuwu:
### General Info

- [X] I checked for similar bug report
- [X] I am using the latest version
- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.MD) page for similar problems

### Version(s)

2.0.0

### Kotlin Target(s) and their respective versions

JVM 17, Android 13

### What happened? (include your code)

Use method `suspend fun upload(path: String, data: ByteArray, upsert: Boolean = false)` from Storage, when empty data is passed, an empty file with 0 byte in size is created on Dashboard as screenshot

<img width="720" alt="Screenshot 2024-04-30 at 20 26 56" src="https://github.com/supabase-community/supabase-kt/assets/43868345/02ff48a8-80ce-4eee-801c-04fd5bce190a">

### Steps To Reproduce (optional)

1. Set up Storage
2. Call `suspend fun upload(path: String, data: ByteArray, upsert: Boolean = false)`  with empty `data` value
3. Open Dashboard of monitored bucket

PR #577, opened 2024-04-30
Avoid empty file created on Dashboard by early checking for empty data

hieuwu:
This is to fix https://github.com/supabase-community/supabase-kt/issues/576

Silently create an empty file as detailed in the issue

## What is the new behavior?
Throw IllegalArgumentException when pass empty data to any upload function requiring data with `ByteArray` type

Screenshot 2024-04-30 at 20 37 32


jan-tennert (approved):
Thanks!

PR #600, opened 2024-05-19
Store session before emitting Authenticated state

iruizmar:

Bug fix

The Authenticated auth status is emitted before persisting the session in storage.

## What is the new behavior?

The session is stored before emitting the Authenticated status.

## Additional context

https://kotlinlang.slack.com/archives/C06QXPC7064/p1716143568376979?thread_ts=1716132763.731959&cid=C06QXPC7064


jan-tennert (approved):
Thanks!

PR #610, opened 2024-05-27
Fix session expiry calculation

JOsacky:

Bug fix

The current behavior calculates the delay before session expiry based on the time between the current system time and the expirationAt time. This delayBeforeExpiry time will change depending on the current system time.

## What is the new behavior?

The new behavior calculates the delay before session expiry based on the time between the beginning of the session and the expirationAt time. This makes it so that the delayBeforeExpiry is static and will always be 80% of the way from the beginning of the session to the expiresAt time.

## Additional context

I'm having some issues with expired tokens being used with my backend from the kotlin client, I'm hoping this helps alleviate some of that

### DIFF
--- a/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AuthImpl.kt
+++ b/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AuthImpl.kt
@@ -50,6 +50,7 @@ import kotlinx.serialization.json.encodeToJsonElement
 import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.put
 import kotlin.math.floor
+import kotlin.time.Duration.Companion.seconds
 
 @PublishedApi
 internal class AuthImpl(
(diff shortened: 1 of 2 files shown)


jan-tennert (approved):
Thanks!

jan-tennert:
@JOsacky I found a relatively major bug which causes sessions to be refreshed later than they should be. Might also cause your issues. A fix is on the way.

JOsacky:
> @JOsacky I found a relatively major bug which causes sessions to be refreshed later than they should be. Might also cause your issues. A fix is on the way.

@jan-tennert thank you so much, I was seeing issues with refreshes happening later. I hope this fix helps, was this the commit you are referring to? 

https://github.com/supabase-community/supabase-kt/commit/9b73c03686f582d9069777b97caccfd006cc7c2f

jan-tennert:
> > @JOsacky I found a relatively major bug which causes sessions to be refreshed later than they should be. Might also cause your issues. A fix is on the way.
> 
> @jan-tennert thank you so much, I was seeing issues with refreshes happening later. I hope this fix helps, was this the commit you are referring to?
> 
> [9b73c03](https://github.com/supabase-community/supabase-kt/commit/9b73c03686f582d9069777b97caccfd006cc7c2f)
> 
> Also thank you so much for coming back to this PR and letting me know!

Yes, let me know if you still get these issues after the fix.

JOsacky:
> > > @JOsacky I found a relatively major bug which causes sessions to be refreshed later than they should be. Might also cause your issues. A fix is on the way.
> > 
> > 
> > @jan-tennert thank you so much, I was seeing issues with refreshes happening later. I hope this fix helps, was this the commit you are referring to?
> 
> Yes, let me know if you still get these issues after the fix.

Awesome, I will pull in the fix now and test

PR #614, opened 2024-06-01
Add the possibility to have a multi-column PK in realtime

iruizmar:

Feature

The realtime API doesn't allow to have a multi-column PK.

## What is the new behavior?

The realtime API now works with multi-column PKs.

## Additional context

https://kotlinlang.slack.com/archives/C06QXPC7064/p1716652456870609

I couldn't really test this. What's the usual strategy to test the library?

### DIFF
--- a/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/PostgrestExtensions.kt
+++ b/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/PostgrestExtensions.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.onCompletion
+import kotlin.jvm.JvmName
 import kotlin.reflect.KProperty1
 
 /**
@@ -123,6 +124,7 @@ inline fun <reified Data : Any, Value> PostgrestQueryBuilder.selectAsFlow(
  * @param channelName the name of the channel to use for the realtime updates. If null, a channel name following the format "schema:table:id" will be used
  */
 @SupabaseExperimental
+@JvmName("selectAsFlowMultiplePks")
 inline fun <reified Data : Any, Value> PostgrestQueryBuilder.selectAsFlow(
     primaryKeys: List<KProperty1<Data, Value>>,
     channelName: String? = null,
--- a/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/PostgrestExtensions.kt
+++ b/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/PostgrestExtensions.kt
@@ -22,20 +22,6 @@ inline fun <reified Data : Any> PostgrestQueryBuilder.selectSingleValueAsFlow(
     primaryKey: PrimaryKey<Data>,
     channelName: String? = null,
     crossinline filter: PostgrestFilterBuilder.() -> Unit
-): Flow<Data> = selectSingleValueAsFlow(listOf(primaryKey), channelName, filter)
-
-/**
- * Executes vertical filtering with select on [PostgrestQueryBuilder.table] and [PostgrestQueryBuilder.schema] and returns a [Flow] of a single value matching the [filter].
- * This function listens for changes in the table and emits the new value whenever a change occurs.
- * @param primaryKeys the list of primary key of the [Data] type
- * @param filter the filter to apply to the select query
- * @param channelName the name of the channel to use for the realtime updates. If null, a channel name following the format "schema:table:id" will be used
- */
-@SupabaseExperimental
-inline fun <reified Data : Any> PostgrestQueryBuilder.selectSingleValueAsFlow(
-    primaryKeys: List<PrimaryKey<Data>>,
-    channelName: String? = null,
-    crossinline filter: PostgrestFilterBuilder.() -> Unit
 ): Flow<Data> {
     val realtime = postgrest.supabaseClient.realtime as RealtimeImpl
     val channel = realtime.channel(channelName ?: defaultChannelName(schema, table, realtime))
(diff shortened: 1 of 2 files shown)


Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeExt.kt
```diff
@@ -21,6 +21,12 @@ import kotlin.reflect.KProperty1
  */
 data class PrimaryKey<Data>(val columnName: String, val producer: (Data) -> String)
 
+fun <Data> List<PrimaryKey<Data>>.producer(data: Data): String =
+    fold("") { value, pk -> value + pk.producer(data) }
+
+val <Data> List<PrimaryKey<Data>>.columnName: String
+    get() = fold("") { value, pk -> value + pk.columnName }
+
```
iruizmar, line 29: This is the key logic change. LMK if this works for you.

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeExt.kt
```diff
@@ -87,20 +113,23 @@ inline fun <reified Data : Any> RealtimeChannel.postgresListDataFlow(
...
                     val data = it.decodeRecord<Data>()
-                    val key = primaryKey.producer(data)
+                    val key = primaryKeys.producer(data)
                     cache[key] = data
                 }
+
                 is PostgresAction.Delete -> {
                     cache.remove(
-                        it.oldRecord[primaryKey.columnName]?.jsonPrimitive?.content
+                        it.oldRecord[primaryKeys.columnName]?.jsonPrimitive?.content
```
jan-tennert, line 128: This can't work with multiple primary keys (the object you receive obviously hasn't a weird concatenated string as key)

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeExt.kt
```diff
@@ -157,13 +203,13 @@ suspend inline fun <reified Data : Any> RealtimeChannel.postgresSingleDataFlow(
...
         val data = result.decodeAs<Data>()
-        primaryKey.producer(data) to data
+        primaryKeys.producer(data) to data
     } catch (e: UnknownRestException) {
         error("Data matching filter and table name $table not found")
     }
     val changeFlow = postgresChangeFlow<PostgresAction>(schema) {
         this.table = table
-        filter(primaryKey.columnName, FilterOperator.EQ, key)
+        filter(primaryKeys.columnName, FilterOperator.EQ, key)
```
jan-tennert, line 212: This can't work either

jan-tennert (changes requested):
Okay, it seems this might take more changes than expected. You can fix the `selectAsFlow` method by just taking the primary key values and combining them yourself. Something like this:
```kotlin
cache.remove(primaryKeys.map { key -> it.oldRecord[key.columnName]?.jsonPrimitive?.content }.joinToString(""))
```
(not sure if that works exactly).
For the single flow variant, I'm not sure if that's possible as we can only provide one filter for realtime

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeExt.kt
iruizmar: Ah, that makes sense. Thanks for the fix.

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeExt.kt
iruizmar: You're right, there's some work in progress apparently: https://github.com/supabase/realtime-js/issues/97

jan-tennert: Yea, I think providing multiple PKs doesn't work for the single value flow for now

iruizmar:
> Okay, it seems this might take more changes than expected. You can fix the `selectAsFlow` method by just taking the primary key values and combining them yourself. Something like this:
> 
> ```kotlin
> cache.remove(primaryKeys.map { key -> it.oldRecord[key.columnName]?.jsonPrimitive?.content }.joinToString(""))
> ```
> 
> (not sure if that works exactly). For the single flow variant, I'm not sure if that's possible as we can only provide one filter for realtime

I made the changes. Can you test that again? 🙏🏻

jan-tennert (commented):
Yes, I think that should work. Can you add `List<KProperty1<Data, Value>>` overloads (like the other overload for `PrimaryKey`, so you can just use this syntax: `selectAsFlow(listOf(Message::id, Message::channelId))`

iruizmar:
> Yes, I think that should work. Can you add `List<KProperty1<Data, Value>>` overloads (like the other overload for `PrimaryKey`, so you can just use this syntax: `selectAsFlow(listOf(Message::id, Message::channelId))`

Oh, I don't know why I missed those. Done ✅.

jan-tennert:
Also, I normally just use the blank `test` module which is included in the gradle project (but in the gitignore, so not in the repo).
Alternatively, you can also publish the dependencies to maven local

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/PostgrestExtensions.kt
```diff
@@ -94,7 +112,27 @@ inline fun <reified Data : Any, Value> PostgrestQueryBuilder.selectAsFlow(
...
+/**
+ * Executes vertical filtering with select on [PostgrestQueryBuilder.table] and [PostgrestQueryBuilder.schema] and returns a [Flow] of a list of values matching the [filter].
+ * This function listens for changes in the table and emits the new list whenever a change occurs.
+ * @param primaryKeys the list of primary key of the [Data] type
+ * @param filter the filter to apply to the select query
+ * @param channelName the name of the channel to use for the realtime updates. If null, a channel name following the format "schema:table:id" will be used
+ */
+@SupabaseExperimental
+inline fun <reified Data : Any, Value> PostgrestQueryBuilder.selectAsFlow(
+    primaryKeys: List<KProperty1<Data, Value>>,
```
jan-tennert, line 127: The JVM test fails because of the `List<>` parameter type (in the bytecode the types doesn't matter, so the JVM thinks there are two methods with essentially the same parameters). Might need to add a `@JvmName` annotation here to fix that

iruizmar: Thanks for that, didn't know how to fix it :)

jan-tennert (approved):
Great! I'm going to include this in the upcoming 2.5.0 beta, so you can try it out yourself!

PR #632, opened 2024-06-25
Added caching to Gradle steps in GitHub Actions

Mithrandir21:

Feature: This PR will add a caching and caching retrieval step to all GitHub Actions steps that involve Gradle.
Caches will make builds faster by not requiring Gradle to download and configure all dependencies for each build.
The cache 'hit' and 'miss' logic is robust and should work well across builds.

No caching is currently implemented in the GitHub Actions steps where Gradle is involved.

## What is the new behavior?

Caches will automatically be created in the GitHub Actions 'caches' section, that will be attempted reused during builds.

## Additional context

Link to GitHub Actions repo:
[Gradle-cache-action](https://github.com/burrunan/gradle-cache-action)

Additionally, there are a lot of settings that can be configurated if and when needed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,8 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
+      - uses: burrunan/gradle-cache-action@v1
+        name: Cache .gradle
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v3.1.0
         with:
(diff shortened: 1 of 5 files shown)


jan-tennert:
Hey, thank you for this PR! We already have Gradle Caching via the `setup-gradle` plugin:
image
https://github.com/supabase-community/supabase-kt/blob/b74002906f9867da8c09a01a46f2fa49bb35baef/.github/workflows/build.yml#L24-L28
Currently, the master branch writes to the cache and all PRs and other branches read from the cache.
If you have any suggestions or improvements regarding the current approach, feel free to leave your feedback!

Issue #573, opened 2024-04-27
[Question]: I think it's better to move the samples to root project?


MohamedRejeb:
### General info

- [X] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.MD) page for similar problems

### What is your question?

HI,
I think it's better to move samples to the root project, it's easier for testing and debugging. You can directly run the sample with the latest changes without the need to publish the library. Also, when you clone the project, the IDE will only recognize the root project and you need to open the samples as separate projects.

jan-tennert:
Yea, I think that should be fine. Originally there were some problems with Compose in the same class path, but that has been fixed a while ago.

jan-tennert:
Handling this currently, unsure if this as easy as it sounds. Each sample has their own modules & settings, which makes it hard to include them all without a lot of boilerplate. Looking at the KotlinX dependencies, most of them have samples with their own gradle project.

jan-tennert:
@MohamedRejeb any idea on this?

MohamedRejeb:
I think that the library needs convention plugins, it will make this much easier.
The build.gradle files will look like this 
https://github.com/MohamedRejeb/Calf/blob/main/calf-file-picker/build.gradle.kts

I can work on this if it makes sense to you.
For modules with different targets we can have different function to setup targets depending on the case.

jan-tennert:
If we mean the same thing, we already have some methods to make the configuration easier:
https://github.com/supabase-community/supabase-kt/blob/master/Storage/build.gradle.kts
But yea, we can probably continue improving on these, if we include all samples. I'm wondering, how should we handle the project build (when changes get merged into master)? We could try to build & compile all samples in the GH actions, but that is probably going to take ages since building the root project already takes 40 min. Same with the PR actions.

jan-tennert:
> root project already takes 40 min.
Probably a bit less if the Gradle files are cached, but its still ~25-30min.

MohamedRejeb:
We can conditionally exclude samples from build when needed. Something like this should work:

```kotlin
if (System.getProperty("LibrariesOnly") != "true") {
    include(":sample:kmp")
}
```

```bash
./gradlew -PLibrariesOnly=true build
```

jan-tennert:
We can close this issue now, all non-deprecated samples have been migrated.

PR #636, opened 2024-06-30
Move chat demo to root project & improve chat demo

jan-tennert:

Samples

The chat demo has their own Gradle instance and is not part of the project.

## What is the new behavior?

The chat demo is now part of the root project and can be used to try out changes easily

## Additional context

See #573


jan-tennert:
@MohamedRejeb thoughts?

sample/chat-demo-mpp/desktop/build.gradle.kts
```diff
@@ -1,25 +1,25 @@
-import org.jetbrains.compose.compose
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 
 plugins {
-    kotlin("multiplatform")
-    id("org.jetbrains.compose")
-    kotlin("plugin.serialization")
+    id(libs.plugins.kotlin.multiplatform.get().pluginId)
```
MohamedRejeb, line 4: This can be the same way as other plugins `alias(libs.plugins.kotlin.multiplatform)`

MohamedRejeb (approved):
LGTM!
For Kotlin GP and AGP, they can be applied with alias the same way as other plugins:
`alias(libs.plugins.kotlin.multiplatform)`

sample/chat-demo-mpp/desktop/build.gradle.kts
jan-tennert: Right thanks, forgot that

jan-tennert: Actually, I remembered I can't use it that way, because buildSrc is making some problems if you use the alias and include Kotlin as a dependency

MohamedRejeb: I see, maybe buildSrc can be changed in the future, it's not a best practice anymore haha

https://github.com/Kotlin/multiplatform-library-template/tree/main/convention-plugins

MohamedRejeb (approved):
LGTM!

PR #642, opened 2024-07-05
fix(realtime): add private to Realtime

grdsdev:

Add support for specifying if channel is private


Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeChannelBuilder.kt
```diff
@@ -11,6 +11,7 @@ class RealtimeChannelBuilder @PublishedApi internal constructor(private val topi
 
     private var broadcastJoinConfig = BroadcastJoinConfig(acknowledgeBroadcasts = false, receiveOwnBroadcasts = false)
     private var presenceJoinConfig = PresenceJoinConfig("")
+    private var isPrivate = false
```
jan-tennert, line 14: I think we can leave this one public, as normally if we have primitive properties in a DSL builder, we don't have extra methods to simplify the syntax (like with the joinConfigs). So this should be just fine:
```kotlin
supabase.realtime.channel {
    isPrivate = true
}

grdsdev: I was in doubt about this, thanks.

jan-tennert (approved):

PR #648, opened 2024-07-08
Fix Google sign in prompt in composeAuth

### LINKED ISSUE #643: [Feature request]: Use GetSignInWithGoogleOption instead of GetGoogleIdOption in ComposeAuth
### General info

- [x] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.md) page for similar problems

### What is your question?

The ComposeAuth plugin uses GetGoogleIdOption for native google sign in.

As per the credential manager documentation [here](https://developer.android.com/identity/sign-in/credential-manager-siwg),

It is advised to use GetSignInWithGoogleOption while signing in with a button.

The GetGoogleIdOption shows a bottom sheet while GetSignInWithGoogleOption shows a dialog.

The GetGoogleIdOption is not meant to use with a button based UI because of a limitation.

Screenshot_20240616-211629

If this setting is disabled on any one of the users google account on the device, the bottomsheet will not appear at all.
(Settings -> Google -> All Services -> Settings for Google apps -> Connected apps -> Third-party apps and services (the gear icon))

Hence, the GetGoogleIdOption method needs to be changed with GetSignInWithGoogleOption which will now show a dialog instead of a bottomsheet.

I am willing to fix this myself, do I need to fork the entire repository even if I want to work on the ComposeAuth directory only?

Sorry, this is my first time contributing 😅 also don't know if this was supposed to be tagged as a bug or a question.

_No response_

Aaron246:

Bug fix #643

The google sign in prompt doesn't show at all if "sign in prompts" setting is disabled in the user's google account/device.

## What is the new behavior?

Sign in prompt (dialog) now shows irrespective of the "sign in prompts" setting is enabled or not.

### DIFF
--- a/plugins/ComposeAuth/src/androidMain/kotlin/io/github/jan/supabase/compose/auth/Utils.kt
+++ b/plugins/ComposeAuth/src/androidMain/kotlin/io/github/jan/supabase/compose/auth/Utils.kt
@@ -9,7 +9,7 @@ internal fun getSignInWithGoogleOptions(
     config: GoogleLoginConfig?,
     nonce: String? = null
 ): GetSignInWithGoogleOption {
-    val signInWithGoogleOption = GetSignInWithGoogleOption.Builder(config!!.serverClientId)
+    val signInWithGoogleOption = GetSignInWithGoogleOption.Builder(config?.serverClientId ?: error("Trying to use Google Auth without setting the serverClientId"))
     signInWithGoogleOption.setNonce(nonce)
     return signInWithGoogleOption.build()
 }
--- a/plugins/ComposeAuth/src/androidMain/kotlin/io/github/jan/supabase/compose/auth/Utils.kt
+++ b/plugins/ComposeAuth/src/androidMain/kotlin/io/github/jan/supabase/compose/auth/Utils.kt
@@ -3,24 +3,15 @@ package io.github.jan.supabase.compose.auth
 import android.app.Activity
 import android.content.Context
 import android.content.ContextWrapper
-import com.google.android.libraries.identity.googleid.GetGoogleIdOption
+import com.google.android.libraries.identity.googleid.GetSignInWithGoogleOption
 
-internal fun getGoogleIDOptions(
+internal fun getSignInWithGoogleOptions(
     config: GoogleLoginConfig?,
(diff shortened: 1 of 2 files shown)


Aaron246:
Hey tan, I believe this is a bug fix and not a enhancement/feature request since the current implementation depends on the user setting "sign in prompts" (details mentioned in the linked issue) which if disabled, the google sign in doesn't work at all. The new implementation proposed by me won't face the same bug since it isn't dependent on the "sign in prompts" user setting.

jan-tennert:
> Hey Jan, I believe this is a bug fix and not a enhancement/feature request since the current implementation depends on the user setting "sign in prompts" (details mentioned in the linked issue) which if disabled, the google sign in doesn't work at all. The new implementation proposed by me won't face the same bug since it isn't dependent on the "sign in prompts" user setting.

Yea, well, in any case, thanks for the PR!

plugins/ComposeAuth/src/androidMain/kotlin/io/github/jan/supabase/compose/auth/Utils.kt
```diff
@@ -3,24 +3,15 @@ package io.github.jan.supabase.compose.auth
...
-        googleIdOption.setFilterByAuthorizedAccounts(filterByAuthorizedAccounts)
-        googleIdOption.setNonce(nonce)
-
-        options.associateLinkedAccounts?.let {
-            googleIdOption.associateLinkedAccounts(it.first, it.second)
-        }
-    }
-    return googleIdOption.build()
+): GetSignInWithGoogleOption {
+    val signInWithGoogleOption = GetSignInWithGoogleOption.Builder(config!!.serverClientId)
```
jan-tennert, line 12:
```suggestion
    val signInWithGoogleOption = GetSignInWithGoogleOption.Builder(config?.serverClientId ?: error("Trying to use Google Auth without setting the serverClientId"))
```
Avoid using `!!`

jan-tennert (commented):
One small thing, otherwise looks good! Did you try out the change?

Aaron246:
> One small thing, otherwise looks good! Did you try out the change?

I did, by publishing it to local Maven repository the onResult successfully return NativeSignInResult.Success.

Issue #447, opened 2024-01-30
[Bug]: Apple Auth is broken


jan-tennert:
> thanks @jan-tennert , I was able to run it locally and have found out where the issue is happening, still not sure if it is something on my end but here is what I found

and Should I create a new issue with these details?

I have added the code and the logs output below

I put some logs in  the method "rememberSignInWithApple" and here are my obervations

1.The method runs as soon as the app opens, without user doing anything, although the code inside launched effect doesnt runs and directly the last log is printed

2. Then when I click on the sign in button all the statments that are marked "executed" run in an instant without me doing anything

3.Nothing is executed after I click on "signIn"(after putting password in the box) in the apple native signin ui

4."nonce" and the "user" properties are showing null

5.All properties inside the config.loginConfig["apple"] object are null, even though object is not null

6.IMPORTANT -looks like authorizationController() method  is never executed, I had put some logs in that method too but nothing was printed and also the print statement "I am here 10" was not printed

```kotlin
@Composable
actual fun ComposeAuth.rememberSignInWithApple(
    onResult: (NativeSignInResult) -> Unit,
    fallback: suspend () -> Unit
): NativeSignInState {

    val state = remember { NativeSignInState() }
    val scope = rememberCoroutineScope()

    LaunchedEffect(key1 = state.started) {
        println("I am inside launched effect")//executed
        if (state.started) {
            println("I am inside first if block , which means state is set to started")//executed

            println("${config.loginConfig["apple"]}")
            if (config.loginConfig["apple"] == null) {
                println("I am inside 2nd if block , which shouldn't happen")// NOT EXECUTED
                fallback.invoke()
                state.reset()
                return@LaunchedEffect
            }

            val appleIDProvider = ASAuthorizationAppleIDProvider()
            val request = appleIDProvider.createRequest().apply {
                requestedScopes = listOf(ASAuthorizationScopeFullName, ASAuthorizationScopeEmail)
                nonce = (config.loginConfig["apple"] as? AppleLoginConfig)?.nonce
                println("loginConfig - ${config.loginConfig["apple"]}  server client id - ${config.loginConfig["apple"]?.serverClientId}")//executed
                println("requested scopes - $requestedScopes , nounce - $nonce  ")//executed
            }
            println("request - $request - user - ${request.user}")//executed


            val controller = ASAuthorizationController(listOf(request)).apply {
                println("i m inside controller ${this.delegate}")//executed
                delegate = authorizationController(scope) {
[...]

PR #654, opened 2024-07-08
fix(auth): fix sign in with Apple in ComposeAuth

grdsdev:

Fix https://github.com/supabase-community/supabase-kt/issues/447

The current code was trying to decode the identity token as `base64` string, but that isn't true. Because of that, it wasn't getting into the last `let` as the `idToken` was null.

## What is the new behavior?

Decodes identityToken as a UTF-8 string

## Additional context

https://github.com/supabase-community/supabase-kt/assets/5923044/e8df4dc1-0528-4509-9a56-fb23be7a7e03


plugins/ComposeAuth/src/appleMain/kotlin/io/github/jan/supabase/compose/auth/composable/AppleAuth.kt
```diff
@@ -82,6 +85,7 @@ actual fun ComposeAuth.rememberSignInWithApple(
     return state
 }
 
+@BetaInteropApi
```
grdsdev, line 88: @jan-tennert I don't know what this is, but I had to add it for using the `NSString.create` method.

Let me know if there is a better option.

jan-tennert: That should be fine

jan-tennert (approved):

PR #664, opened 2024-07-17
Introduce RESTful service built with supabase-kt

hieuwu:
Introduce RESTful service sample built with supabase-kt
https://github.com/hieuwu/supa-spring-kt

Introduce RESTful service built with supabase-kt

## What is the new behavior?
N/A

## Additional context
N/A


jan-tennert (approved):
We will probably have to improve this sample list sometime to be clearer, but that's for another PR!

PR #688, opened 2024-08-12
Add support for third party auth and add Slack OIDC provider

jan-tennert:

Feature

No support for third party auth and missing the Slack OIDC provider.

## What is the new behavior?

The provider has been added and there is a new option in the `SupabaseClientBuilder`:
```kotlin
val supabase = createSupabaseClient(supabaseUrl, supabaseKey) {
    accessToken = {
          //fetch the third party token
         "my-token"
    }
}
```
If this property is not null, the `Auth` plugin will not be available and this custom provider function will be used for auth.

## Additional context

Check https://github.com/supabase/supabase-js/pull/1004.

### DIFF
--- a/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AccessToken.kt
+++ b/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AccessToken.kt
@@ -1,0 +1,36 @@
+package io.github.jan.supabase.gotrue
+
+import io.github.jan.supabase.SupabaseClient
+import io.github.jan.supabase.annotations.SupabaseInternal
+import io.github.jan.supabase.plugins.MainConfig
+import io.github.jan.supabase.plugins.MainPlugin
+
+/**
+ * Returns the access token used for requests. The token is resolved in the following order:
+ * 1. [jwtToken] if not null
+ * 2. [SupabaseClient.customAccessToken] if not null
+ * 3. [Auth.currentAccessTokenOrNull] if the Auth plugin is installed
+ * 4. [SupabaseClient.supabaseKey] if [keyAsFallback] is true
+ */
+@SupabaseInternal
+suspend fun SupabaseClient.accessToken(
+    jwtToken: String? = null,
+    keyAsFallback: Boolean = true
+): String? {
+    val key = if(keyAsFallback) supabaseKey else null
+    return jwtToken ?: customAccessToken?.invoke()
+    ?: pluginManager.getPluginOrNull(Auth)?.currentAccessTokenOrNull() ?: key
+}
+
+/**
+ * Returns the access token used for requests. The token is resolved in the following order:
+ * 1. [jwtToken] if not null
+ * 2. [SupabaseClient.customAccessToken] if not null
+ * 3. [Auth.currentAccessTokenOrNull] if the Auth plugin is installed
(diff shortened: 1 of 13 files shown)


jan-tennert:
Does this also apply to realtime? @grdsdev @dshukertjr

Supabase/src/commonMain/kotlin/io/github/jan/supabase/AccessTokenProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Supabase. The function should return an access token or ID token (JWT) by
+ * obtaining it from the third-party auth client library. Note that this
+ * function may be called concurrently and many times. Use memoization and
+ * locking techniques if this is not supported by the client libraries.
+ *
+ * When set, the Auth plugin from `auth-kt` cannot be used.
+ * Create another client if you wish to use Supabase Auth and third-party
+ * authentications concurrently in the same application.
+ */
+typealias AccessTokenProvider = () -> String
```
grdsdev, line 14: This should be async, as some of the 3p providers returns the token asynchronously.

grdsdev (changes requested):

GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AccessToken.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Returns the access token used for requests. The token is resolved in the following order:
+ * 1. [jwtToken] if not null
+ * 2. [SupabaseClient.customAccessToken] if not null
+ * 3. [Auth.currentAccessTokenOrNull] if the Auth plugin is installed
+ * 4. [SupabaseClient.supabaseKey] if [keyAsFallback] is true
+ */
+@SupabaseInternal
+suspend fun SupabaseClient.accessToken(
+    jwtToken: String? = null,
+    keyAsFallback: Boolean = true
```
dshukertjr, line 18: Do we need these two parameters? They are not present on the js lib, right?

dshukertjr (commented):
Sorry for the delay, but yes this should work with all services including realtime! Thanks for working on this!

GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AccessToken.kt
jan-tennert: This method is just an internal utility function, hence the `@SupabaseInternal` annotation. The actual public api is exactly the same as JS (located in the SupabaseClientBuilder):
```kotlin
val supabase = createSupabaseClient(supabaseUrl, supabaseKey) {
    accessToken = {
          //fetch the third party token
         "my-token"
    }
}
```

jan-tennert: And this method can be accessed via `supabase.customAccessToken()`, but we can rename this if this fits better.

dshukertjr: > This method is just an internal utility function

Oh yeah, what was I looking at 😂

> we can rename this if this fits better.

I think it's fine to just keep it in!

dshukertjr (approved):

PR #696, opened 2024-08-22
Refactor Realtime event system

jan-tennert:

Internal refactor

Currently, the event types are parsed through one complex property (`RealtimeMessage.EventType`) and handled in only one method (`onMessage`).
This makes the code very complex and hard to see through.

## What is the new behavior?

There are now `RealtimeEvent`s, which implement two methods: `handle` and `appliesTo`. The `appliesTo` method is used to check whether the `handle` method should be called for this event.
All former `EventType`s now have their own `RealtimeEvent` implementation.
This simplifies the event code and makes it way easier to extend.

### DIFF
--- a/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/event/RealtimeEvent.kt
+++ b/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/event/RealtimeEvent.kt
@@ -13,7 +13,7 @@ internal sealed interface RealtimeEvent {
      * @param channel The channel the event was received on.
      * @param message The message that was received.
      */
-    fun handle(channel: RealtimeChannel, message: RealtimeMessage)
+    suspend fun handle(channel: RealtimeChannel, message: RealtimeMessage)
 
     /**
      * Checks if the event applies to the message.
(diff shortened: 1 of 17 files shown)


Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/event/RealtimeEvent.kt
```diff
@@ -0,0 +1,47 @@
+package io.github.jan.supabase.realtime.event
+
+import io.github.jan.supabase.realtime.RealtimeChannel
+import io.github.jan.supabase.realtime.RealtimeMessage
+
+/**
+ * Interface for handling realtime events.
+ */
+internal sealed interface RealtimeEvent {
```
jan-tennert, line 9: We could possibly rename this and subclasses to `RealtimeEventHandler`

dshukertjr: I'm fine either way since it's an internal symbol 👍

dshukertjr (approved):

jan-tennert:
@grdsdev I'd merge this for know as this is definitely an improvement on the current system. We can still go on from there.

PR #716, opened 2024-09-08
Extract optional PostgREST properties and add schema support for RPC.

### LINKED ISSUE #713: Add `schema` option for database functions
### Discussed in https://github.com/supabase-community/supabase-kt/discussions/712

<div type='discussions-op-text'>

<sup>Originally posted by **Premjit-Developer** September  4, 2024</sup>
I'm trying to create a custom schema and have added some functions to it, but I'm unsure how to call these functions properly. This is because rpc does not have schema parameter like get() or from().</div>

There is currently no easy way to specify the DB schema for calling a DB function. As a workaround, you can set the header yourself:
```kotlin
postgrest.rpc("my_function") {
   headers["Content-Profile"] = "my_schema"
} 
```

jan-tennert:

Feature (closes #713)

Optional properties like `defaultToNull` for inserting or `head` for selecting are a function parameter. In increasing numbers, these function parameters get too clunky and as there are already some parameters in the DSL builder (`PostgrestRequestBuilder`), moving these properties would fit more.

## What is the new behavior?

The Select, Insert and Upsert methods now have their own DSL inheriting from the base builder and adding these "old" function parameters. The syntax now looks like this:
```kotlin
supabase.from("table").select {
    head = true
}
supabase.from("table").upsert(myValue) {
    defaultToNull = false
    ignoreDuplicates = false
}
supabase.postgrest.rpc("function") {
    method = RpcMethod.GET
    schema = "mySchema"
}
```

## Additional context

### DIFF
--- a/Postgrest/src/commonTest/kotlin/PostgrestRequestBuilderTest.kt
+++ b/Postgrest/src/commonTest/kotlin/PostgrestRequestBuilderTest.kt
@@ -1,7 +1,8 @@
 import io.github.jan.supabase.postgrest.query.Count
 import io.github.jan.supabase.postgrest.query.Order
 import io.github.jan.supabase.postgrest.query.Returning
-import io.ktor.http.*
+import io.github.jan.supabase.postgrest.query.postgrestRequest
+import io.ktor.http.HttpHeaders
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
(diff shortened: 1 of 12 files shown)


dshukertjr (approved):

Postgrest/src/commonTest/kotlin/PostgrestRequestBuilderTest.kt
```diff
@@ -1,8 +1,7 @@
 import io.github.jan.supabase.postgrest.query.Count
 import io.github.jan.supabase.postgrest.query.Order
 import io.github.jan.supabase.postgrest.query.Returning
-import io.github.jan.supabase.postgrest.query.postgrestRequest
-import io.ktor.http.HttpHeaders
+import io.ktor.http.*
```
jan-tennert, line 4:
```suggestion
import io.github.jan.supabase.postgrest.query.postgrestRequest
import io.ktor.http.HttpHeaders
```

PR #694, opened 2024-08-21
Add support for file metadata, `info` and `exists`

jan-tennert:
## What kind of change does this PR introduce?

Feature

## What is the new behavior?

Support for the new user metadata and the new `exists` and `info` methods.
```kotlin
client.storage["my-bucket"].upload("file.png", data) {
    userMetadata {
        put("key", "value")
    }
    //Or:
    userMetadata(myCustomObject)
}
```
Additionally, the `Storage` plugin now also provides the ability to set a custom serializer.

## Additional context

See https://github.com/supabase/storage-js/pull/207/

Issue #415, opened 2024-01-05
[Feature request]: package to use jwt for backend


landsman:
### General Info

- [X] I installed the latest version of Supabase Kt
- [X] I checked for similar feature requests

### Feature request

Please provide a package to decode JWT auth. Including Supabase user interface etc.
It would be lovely to have ready-made integration for the most common Spring Boot Framework (Java/Kotlin).

Inspiration: https://github.com/tschuehly/htmx-supabase-spring-boot-starter/tree/master/src/main/kotlin/de/tschuehly/htmx/spring/supabase/auth

### Usecase

Typically you have React / Next.js application for frontend and backend in Spring Boot. So you want to get a Supabase access token and verify it on the backend to let the user in and pair it to backend entities.

- https://github.com/tschuehly/htmx-supabase-spring-boot-starter
- https://github.com/tschuehly/demo-supabase-spring-boot-starter

jan-tennert:
Could you clarify what exactly such a library should provide? So JWT decoding and verification?

nateweisz:
> Could you clarify what exactly such a library should provide? So JWT decoding and verification?

I'd also be interested in this for ktor as a server. Our frontend uses supabase for authentication and I need to be able to intercept requests and check for authentication stuff.

jan-tennert:
> > Could you clarify what exactly such a library should provide? So JWT decoding and verification?
> 
> I'd also be interested in this for ktor as a server. Our frontend uses supabase for authentication and I need to be able to intercept requests and perform authentication checks on the backend.

So it should work like a ktor plugin which automatically checks/rejects requests based on the JWT?

nateweisz:
> > > Could you clarify what exactly such a library should provide? So JWT decoding and verification?
> > 
> > 
> > I'd also be interested in this for ktor as a server. Our frontend uses supabase for authentication and I need to be able to intercept requests and perform authentication checks on the backend.
> 
> So it should work like a ktor plugin which automatically checks/rejects requests based on the JWT?

Yes, or at least makes it possibly to build that into your system without too much code.

jan-tennert:
Alright, if anyone wants to try to take this on, feel free to ask me for questions for integrating this into supabase-kt. I think ideally there is some basic JWT stuff inside gotrue-kt and then there are spring-boot/ktor modules inside /plugins (like Compose Auth Ui which is not really a supabase-kt plugin)

landsman:
[...]

PR #729, opened 2024-09-16
Refactor Storage download & upload (flow) implementation

jan-tennert:

Internal refactor and public API improvements

The current implementation is very complex/messy and the flow variants are even worse (calling `BucketApiImpl` methods, that shouldn't be in a public extension function).

### Public API

Every upload and download method now has a `options` DSL. This makes it possible to add configuration options in the future without filling up the parameter list (e.g. file metadata)
For downloads, there are currently only two things to configure:
1. Image transformation (moved down)
2. Http request overrides. This is important for the flow variants to attach download listeners.

Example of the new syntax:
```kotlin
supabase.storage.from("test").downloadAuthenticated("test.jpg") {
    transform {
        size(100, 100)
    }
}
```

For uploads, you can configure:
1. Whether to upsert or not (moved down)
2. The content type (will be inferred if null, wasn't configurable before)
3. Also Http request overrides for flow variants.

Example of the new syntax:
```kotlin
supabase.storage.from("test").upload("test.txt", "Hello World!".encodeToByteArray()) {
    contentType = ContentType.Text.Plain
    upsert = true
}
```

Every upload/update/download variant has this new builder including the flow variants.

`BucketItem` has been renamed to `FileObject` to match the JS lib.

### Internal implementation

For non-flow-variants I cleaned up the code to minimize duplications & improve readability.
The same approach was used for the flow variants, which no longer call internal methods from `BucketApiImpl` and instead use the new `httpOverride` option.

## Additional context


jan-tennert:
Note that there are some commits from #694 because I backported and improved some of the changes, but file metadata and `exists`/`info` support is not being added in this PR.

dshukertjr (approved):
I love the new public API! Just to check, these updates are being included in another major version bump, correct?

jan-tennert:
Yes, I originally wanted to only refactor some internal stuff, but now the changes are getting very breaking alongside the WASM support. I can probably include #415 too.

PR #769, opened 2024-10-30
Introduce KSP compiler for generating columns via `@Selectable` annotation

jan-tennert:

Feature

## What is the new behavior?

New KSP compiler module which processes `@Selectable` annotated data classes. I kept the KSP compiler more general, so if we ever want to add a new annotation, we can just use this module.
### How it works
- [**Add the KSP compiler**](https://github.com/supabase-community/supabase-kt/tree/postgrest-column-ksp/KSP)

- **Annotate your data models with `@Selectable`:**
```kotlin
@Selectable
@Serializable
data class Movie(
    val id: Int,

    val name: String

    @ColumnName("created_at") //if the parameter names differs from the column name, add this annotation
    val createdAt: Instant,

    @Foreign //Mark a foreign relationship
    @ColumnName("studio_id") //Either the table name or column name
    val studio: Studio //This should also be a `@Selectable` type

    @ColumnName("metadata") //query JSON columns
    @JsonPath("video", "resolution") 
    val resolution: String
)
```
- **Call the generated `addSelectableTypes()` method in the PostgrestConfig:**
```kotlin
install(Postgrest) {
    addSelectableTypes()
}
```
- **Use the new overloads for selecting:**
```kotlin
val movies: List<Movie> = supabase.from("movies").select<Movie>().decodeList()
//or for insert, update, etc.:
//Note that `myNewMovie` is obviously a different type
val updatedMovie: Movie = supabase.from("movies").insert(myNewMovie) { 
    select<Movie>()
}
```

### Design details

There are 5 annotations: 
- `@Serializable` - Data classes will be processed by the compiler and columns will be generated based of the constructor parameters.
- `@ColumnName` - Used to specify the actual column name when the parameter name differs. For foreign columns, this can be either the column name or foreign table.
- `@JsonPath` - Used to specify a Json Path to query a JSON column
- `@Foreign` - Used for foreign columns so the compiler knows that it has to generate nested columns
- `@ApplyFunction` - Used to apply an aggregate function

**Todo:**
- [ ] Find a solution for spreading embedded resources 
- [ ] Tests

### Implementation

There is a new Postgrest config property `columnRegistry` which stores column data under the full class name for annotated data classes. The KSP compiler fills this registry when the generated `addSelectableTypes()` method is called.
Example for the generated method **[Not visible normally]**:
```kotlin
/**
 * Adds the types annotated with [Selectable] to the ColumnRegistry. Allows to use the automatically generated columns in the PostgrestQueryBuilder.
 *
 * This file is generated by the SelectableSymbolProcessor.
 * Do not modify it manually.
 */
@OptIn(SupabaseInternal::class)
public fun Postgrest.Config.addSelectableTypes() {
  columnRegistry.registerColumns("Movie", "id,title:movie_title,studio:studios(id,name),someAverage:column_name.avg(),jsonKey:json_data->array->0->>key")
  columnRegistry.registerColumns("Studio", "id,name")
}

```

## Additional context

See discussion in #761

### DIFF
--- a/KSP/src/main/kotlin/io/github/jan/supabase/ksp/SelectableSymbolProcessor.kt
+++ b/KSP/src/main/kotlin/io/github/jan/supabase/ksp/SelectableSymbolProcessor.kt
@@ -9,6 +9,7 @@ import com.google.devtools.ksp.symbol.KSAnnotated
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSNode
 import com.google.devtools.ksp.symbol.KSValueParameter
+import com.google.devtools.ksp.symbol.Modifier
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.PropertySpec
@@ -32,8 +33,8 @@ class SelectableSymbolProcessor(
         symbols.forEach { symbol ->
             val className = symbol.simpleName.asString()
             val packageName = symbol.containingFile?.packageName?.asString().orEmpty()
-            if (!symbol.modifiers.containsIgnoreCase("data")) {
-                logger.error("This object is not a data class", symbol)
+            if (!symbol.modifiers.contains(Modifier.DATA)) {
+                logger.error("The class $className is not a data class", symbol)
                 return emptyList()
             }
             val companionObject = symbol.anyCompanionObject()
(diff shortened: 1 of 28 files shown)


KSP/src/main/kotlin/io/github/jan/supabase/ksp/SelectableSymbolProcessor.kt
```diff
@@ -0,0 +1,232 @@
...
+    val packageName = options["selectablePackageName"] ?: "io.github.jan.supabase.postgrest"
+    val fileName = options["selectableFileName"] ?: "PostgrestColumns"
+
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        val symbols = resolver.getSymbolsWithAnnotation(Selectable::class.java.name).filterIsInstance<KSClassDeclaration>()
+        if (!symbols.iterator().hasNext()) return emptyList()
+        val types = hashMapOf<String, String>()
+        symbols.forEach { symbol ->
+            val className = symbol.simpleName.asString()
+            val qualifiedName = symbol.simpleName.asString() //Kotlin JS doesn't support qualified names
```
jan-tennert, line 42: Name clashes could be possible due to only using the simple name, but we cannot use qualified names because of Kotlin JS support.

grdsdev:
Awesome @jan-tennert let me try to get roughly the same API in Swift before merging this ok? But it looks awesome.

jan-tennert:
@grdsdev Update on this?

Izzaddin97 (changes requested):
****

PR #875, opened 2025-03-16
Fix JS/Wasm hash check 

Sparky983:

Fix

Currently an error is thrown on initialisation since a closure is not a valid state object (it is not cloneable).

image

## What is the new behavior?

The code works as expected.

## Additional context

The WASM JS checks are [implemented correctly](https://github.com/supabase-community/supabase-kt/blob/2beaac5831f3988fe4fc44be32361b5e009748cd/Auth/src/wasmJsMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt#L25).

### DIFF
--- a/Auth/src/jsMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
+++ b/Auth/src/jsMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
@@ -23,7 +23,7 @@ actual fun Auth.setupPlatform() {
 
         parseFragmentAndImportSession(afterHash) {
             val newURL = window.location.href.split("?")[0];
-            window.history.replaceState({}, window.document.title, newURL);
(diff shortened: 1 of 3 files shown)


Auth/src/jsMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
```diff
@@ -23,7 +23,7 @@ actual fun Auth.setupPlatform() {
 
         parseFragmentAndImportSession(afterHash) {
             val newURL = window.location.href.split("?")[0];
```
Sparky983, line 25: I believe this should also be `split("#")` however I don't have time to change it.

jan-tennert: Let me check this

PR #883, opened 2025-03-30
Remove sealed modifier where its not needed

jan-tennert:

Fix

Some interface have the `sealed` modifier, even though there are not needed. This makes it impossible to mock them.

## What is the new behavior?

The modifier has been removed from certain interfaces.


mpierucci:
So quick! Really appreciated!

Issue #889, opened 2025-04-02
[Bug]: CancellationExceptions are not always rethrown


sproctor:
### General Info

- [x] I checked for similar bug report
- [x] I am using the latest version
- [x] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.md) page for similar problems

### Version(s)

3.1.3

### Kotlin Target(s) and their respective versions

All

### What happened? (include your code)

I'm getting logged exceptions being handled from cancelled coroutines. Any try/catch containing a suspend call should check for and rethrow `CancellationExceptions`.

My code triggering this:
```kotlin
coroutineScope {
  val channel = supabase.realtime.channel("db-changes")
  channel.postgresChangeFlow<PostgresAction>(schema = "public") {
              ...
  }
  .onEach { ... }
  .launchIn(this)
  channel.subscribe(true)
  try {
    awaitCancellation()
  } finally {
    withContext(NonCancellable) {
      channel.unsubscribe()
      supabase.realtime.removeChannel(channel)
    }
  }
}
```

### Relevant log output (optional)

Unfortunately, I'm not getting the full stacktrace in my logs.

```shell
ChildCancelledException: Child of the scoped flow was cancelled
```

jan-tennert:
So trying to reproduce this, how exactly is this exception triggered? My repro code:
```kotlin
    coroutineScope {
        val job = launch {
            val channel = supabase.realtime.channel("db-changes")
            channel.postgresChangeFlow<PostgresAction>(schema = "public") {
                table = "messages"
            }
                .onEach { println(it) }
                .launchIn(this)
            channel.subscribe(true)
            try {
                awaitCancellation()
            } finally {
                withContext(NonCancellable) {
                    channel.unsubscribe()
                    supabase.realtime.removeChannel(channel)
                }
            }
        }
        delay(2000)
        job.cancel()
    }
```

sproctor:
I think it only happens when the access token is expired. This `catch` seems to be the problematic one: https://github.com/supabase-community/supabase-kt/blob/c4163bb495ad5d5b8133ad0a3e523fb5bb7a7fd5/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt#L85

sproctor:
Fixed by PR #895

PR #895, opened 2025-04-11
Don't swallow `CancellationException`

sproctor:
Add checks that the coroutine is still active in places that can catch `CancellationException`

Bug fix

#889 

## What is the new behavior?

Throws a `CancellationException` in `Exception` handling logic. See: https://medium.com/better-programming/the-silent-killer-thats-crashing-your-coroutines-9171d1e8f79b and https://github.com/Kotlin/kotlinx.coroutines/issues/3658 for the rationale.

### DIFF
--- a/Supabase/src/commonMain/kotlin/io/github/jan/supabase/Utils.kt
+++ b/Supabase/src/commonMain/kotlin/io/github/jan/supabase/Utils.kt
@@ -61,7 +61,7 @@ suspend inline fun <reified T> HttpResponse.bodyOrNull(): T? {
     return try {
         val text = bodyAsText()
         supabaseJson.decodeFromString<T>(text)
-    } catch(e: Exception) {
+    } catch(_: Exception) {
         null
     }
 }
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/Auth.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/Auth.kt
@@ -25,8 +25,10 @@ import io.github.jan.supabase.plugins.CustomSerializationPlugin
 import io.github.jan.supabase.plugins.MainPlugin
 import io.github.jan.supabase.plugins.SupabasePluginProvider
 import io.ktor.client.plugins.HttpRequestTimeoutException
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.serialization.json.JsonObject
+import kotlin.coroutines.coroutineContext
 
 /**
  * Plugin to interact with the Supabase Auth API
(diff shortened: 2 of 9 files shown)


Supabase/src/commonMain/kotlin/io/github/jan/supabase/Utils.kt
```diff
@@ -61,7 +61,7 @@ suspend inline fun <reified T> HttpResponse.bodyOrNull(): T? {
     return try {
         val text = bodyAsText()
         supabaseJson.decodeFromString<T>(text)
-    } catch(e: Exception) {
+    } catch(_: Exception) {
```
sproctor, line 64: Ensuring the coroutine is active here, causes `BucketApiTest.testExistsWithNonExistingFile` to fail. I think there's a deeper issue in that test that gets exposed. The test shouldn't still be running if the job is completed. I think there might be a missing `await` somewhere there.

jan-tennert:
Were you able to test if this fixes the problem in #889?

Supabase/src/commonMain/kotlin/io/github/jan/supabase/Utils.kt
jan-tennert: Then I'm wondering why only that test fails :thinking:

jan-tennert: Okay there seems to be a problem with this method in general, this also fails with ensure active:
```kotlin
val client = HttpClient()
val res = client.get("https://google.de")
println(res.bodyOrNull<String>())
```

sproctor:
It fixes #889. The overall behavior seems to be better, but I'm not sure if it was these changes or something else. I'm still getting:

```
ClosedReceiveChannelException: Channel was closed
Error while listening for messages. Trying again in 7s
```

I think that might be unavoidable when the access token expires while connected. I guess I could add some logic to close the connection and refresh the token before it expires if this turns out to be problematic.

Thanks for your help and your work on this.

jan-tennert:
> I think that might be unavoidable when the access token expires while connected. I guess I could add some logic to close the connection and refresh the token before it expires if this turns out to be problematic.

Well under normal circumstances, this should not happen as this is already handled:
https://github.com/supabase-community/supabase-kt/blob/c4163bb495ad5d5b8133ad0a3e523fb5bb7a7fd5/Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt#L96-L109

The tokens are refreshed via the Auth plugin 20% before their expiration time and Realtime waits for any refreshes or session changes

jan-tennert (approved):
Thanks!

Supabase/src/commonMain/kotlin/io/github/jan/supabase/Utils.kt
jan-tennert: I'll address this in a separate PR

PR #905, opened 2025-04-17
Use Dispatchers.IO by default on multi-threaded platforms

sproctor:
and allow dispatcher to be configured in realtime and storage

Bug fix/feature

`Dispatchers.Default` is used for all operations, but configurable in `Auth`.

## What is the new behavior?

`Dispatchers.IO` is used by default on JVM and K/N targets, `Dispatchers.Default` is still used on JS and WASM targets. All invocations are configurable by an appropriate config.

## Additional context

`Dispatchers.Default` is intended for CPU bound operations and is limited by the number of CPU cores. IO operations should use `Dispatchers.IO` instead.

### DIFF
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
@@ -51,6 +51,7 @@ open class AuthConfigDefaults : MainConfig() {
     /**
      * The dispatcher used for all auth related network requests
      */
+    @Deprecated("SupabaseClientBuilder.coroutineDispatcher should be used instead")
     var coroutineDispatcher: CoroutineDispatcher? = null
 
     /**
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
@@ -2,11 +2,9 @@ package io.github.jan.supabase.auth
 
 import io.github.jan.supabase.SupabaseClientBuilder
 import io.github.jan.supabase.SupabaseSerializer
-import io.github.jan.supabase.defaultDispatcher
 import io.github.jan.supabase.plugins.CustomSerializationConfig
 import io.github.jan.supabase.plugins.MainConfig
 import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
@@ -53,7 +51,7 @@ open class AuthConfigDefaults : MainConfig() {
     /**
      * The dispatcher used for all auth related network requests
      */
-    var coroutineDispatcher: CoroutineDispatcher = defaultDispatcher
+    var coroutineDispatcher: CoroutineDispatcher? = null
 
     /**
      * The type of login flow to use. Defaults to [FlowType.IMPLICIT]
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
@@ -2,6 +2,7 @@ package io.github.jan.supabase.auth
 
 import io.github.jan.supabase.SupabaseClientBuilder
 import io.github.jan.supabase.SupabaseSerializer
+import io.github.jan.supabase.defaultDispatcher
 import io.github.jan.supabase.plugins.CustomSerializationConfig
 import io.github.jan.supabase.plugins.MainConfig
 import kotlinx.coroutines.CoroutineDispatcher
@@ -52,7 +53,7 @@ open class AuthConfigDefaults : MainConfig() {
     /**
      * The dispatcher used for all auth related network requests
      */
-    var coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default
+    var coroutineDispatcher: CoroutineDispatcher = defaultDispatcher
 
     /**
      * The type of login flow to use. Defaults to [FlowType.IMPLICIT]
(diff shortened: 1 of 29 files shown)


sproctor:
I think it would be cleaner if `defaultDispatcher` were made internal and set in the top level config rather than in each module that uses dispatchers. I'm just not sure how to access that config from the other modules. I guess it can be a property of `SupabaseClient`. Let me know what you think of these changes and if you'd prefer that and what to do about the existing `coroutineDispatcher` property in `Auth`.

jan-tennert:
We could do something similar to the serializers. So you can set a default dispatcher in the `SupabaseClientBuilder`:
```kotlin
val supabase = createSupabaseClient(url, key) {
     defaultDispatcher = ... //By default set to the corresponding platform dispatcher
}
```
.. maybe even `defaultIoDispatcher` so its clear?
Introduce a `CustomDispatcherConfig` similar to that interface:
https://github.com/supabase-community/supabase-kt/blob/4fd0c66812f55c6d093282a64cea615e3a35e8ff/Supabase/src/commonMain/kotlin/io/github/jan/supabase/plugins/CustomSerializationConfig.kt#L6-L16

Then plugins which actually use a scope override this interface in their config, defaulting to null. And in the plugin implementation you have something like this:
```kotlin
val dispatcher = config.ioDispatcher ?: supabase.defaultIoDispatcher //defaultIoDispatcher is obviously non-nullable
val scope = CoroutineScope(dispatcher...)
```

For Auth, if we go with `ioDispatcher` as the config name, you could probably deprecate the `coroutineDispatcher` property and use this in auth:
```kotlin
val dispatcher = config.ioDispatcher ?: config.coroutineDispatcher ?: supabase.defaultIoDispatcher
val scope = CoroutineScope(dispatcher...)
```

sproctor:
Sorry, I missed the notification on the initial comment. I'm not sure there's a ton of value to being able to set the dispatcher per module. It probably makes the most sense to just put the dispatcher in `SupabaseClientBuilder` defaulting to `Dispatchers.IO` on targets that have it and deprecate the config in `Auth`.

sproctor:
Okay, I'll re-work this now

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthImpl.kt
```diff
@@ -72,7 +72,7 @@ internal class AuthImpl(
 
     private val _sessionStatus = MutableStateFlow<SessionStatus>(SessionStatus.Initializing)
     override val sessionStatus: StateFlow<SessionStatus> = _sessionStatus.asStateFlow()
-    internal val authScope = CoroutineScope(config.coroutineDispatcher)
+    internal val authScope = CoroutineScope(config.coroutineDispatcher ?: supabaseClient.coroutineDispatcher)
```
sproctor, line 75: Should this be a supervisor scope? My first instinct is yes.

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
```diff
@@ -52,7 +51,7 @@ open class AuthConfigDefaults : MainConfig() {
     /**
      * The dispatcher used for all auth related network requests
      */
-    var coroutineDispatcher: CoroutineDispatcher = Dispatchers.Default
+    var coroutineDispatcher: CoroutineDispatcher? = null
```
jan-tennert, line 54: Should be deprecated

Storage/src/commonMain/kotlin/io/github/jan/supabase/storage/Storage.kt
```diff
@@ -120,7 +120,7 @@ interface Storage : MainPlugin<Storage.Config>, CustomSerializationPlugin {
     data class Config(
         var transferTimeout: Duration = 120.seconds,
         @PublishedApi internal var resumable: Resumable = Resumable(),
-        override var serializer: SupabaseSerializer? = null
+        override var serializer: SupabaseSerializer? = null,
```
jan-tennert, line 123:
```suggestion
        override var serializer: SupabaseSerializer? = null
```

Storage/src/commonMain/kotlin/io/github/jan/supabase/storage/resumable/ResumableUpload.kt
```diff
@@ -84,7 +84,8 @@ internal class ResumableUploadImpl(
     private val httpClient: HttpClient,
     private val storageApi: BucketApi,
     private val retrieveServerOffset: suspend () -> Long,
-    private val removeFromCache: suspend () -> Unit
+    private val removeFromCache: suspend () -> Unit,
+    coroutineDispatcher: CoroutineDispatcher,
```
jan-tennert, line 88:
```suggestion
    coroutineDispatcher: CoroutineDispatcher
```

Supabase/src/commonMain/kotlin/io/github/jan/supabase/SupabaseClientBuilder.kt
```diff
@@ -69,6 +70,13 @@ class SupabaseClientBuilder @PublishedApi internal constructor(private val supab
      */
     var defaultSerializer: SupabaseSerializer = KotlinXSerializer(Json { ignoreUnknownKeys = true })
 
+    /**
+     * The CoroutineDispatcher used for launching long running jobs.
+     *
+     * Default: [KotlinXSerializer]
```
jan-tennert, line 76:
```suggestion
     * Default: [defaultDispatcher], the IO Dispatcher on supported targets.
```

jan-tennert (changes requested):

Supabase/src/commonMain/kotlin/io/github/jan/supabase/SupabaseClientBuilder.kt
sproctor: whoops

jan-tennert (approved):

PR #912, opened 2025-04-25
Add new event system, add support for error code in OTP links

jan-tennert:

Feature

- You have to do the parsing the error codes in the URL yourself. (If there is one e.g. in deeplinks on Android/iOS, current URL on JS, etc.)
- The `sessionStatus` will get updated to `SessionStatus.RefreshFailure(e)` even if its still valid (=no access to the valid session).

## What is the new behavior?

New event system which exists next to the `sessionStatus`, but works independently. Currently there are two events:
- `AuthEvent.OtpError(...)` - will be emitted if an error code was found in a OTP link (deeplink on Android/iOS, URL on JS and Desktop)
- `AuthEvent.RefreshFailure(cause)` - will be emitted if a session refresh failed (regardless if the session is still valid)

This PR also changes the `SessionStatus.RefreshFailure(cause)`:
- The `cause` parameter/property is deprecated (use the event for the cause)
- This status will only get set, if the session expired. 

--> If a refresh failed, an event will always be emitted, but the session status will only get updated if the session also expired. Planning to rename the status in the future to something like `SessionStatus.NeedsRefresh`

Additional changes:
- Error related parameters will now be removed from the history when used


Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/status/NotAuthenticatedReason.kt
```diff
@@ -0,0 +1,35 @@
+package io.github.jan.supabase.auth.status
+
+import io.github.jan.supabase.auth.exception.AuthErrorCode
+
+/**
+ * Represents the reason for an [SessionStatus.NotAuthenticated] status.
+ */
+sealed interface NotAuthenticatedReason {
```
jan-tennert, line 8: Thinking about removing this entirely and just using the two parameters in the `NotAuthenticated` status: `isSignOut`, `error`

sproctor:
Just a thought, would it make sense to merge `SessionStatus.RefreshFailure` into this with a `RefreshFailure` reason?

jan-tennert:
> Just a thought, would it make sense to merge `SessionStatus.RefreshFailure` into this with a `RefreshFailure` reason?

Maybe. But that would be too breaking for a minor version, we can leave this thought for v4.
I'm not yet happy with this design anyways, so it might be something else entirely.

jan-tennert:
> > Just a thought, would it make sense to merge `SessionStatus.RefreshFailure` into this with a `RefreshFailure` reason?
> 
> Maybe. But that would be too breaking for a minor version, we can leave this thought for v4. I'm not yet happy with this design (NotAuthenticatedReason) anyways, so it might be something else entirely.

Planning to refactor this:
https://github.com/supabase-community/supabase-kt/issues/919#issuecomment-2843097855

dshukertjr (approved):

grdsdev (commented):
Hi @jan-tennert these changes deviates from the official libs, I'd like to better understand its benefits, could you explain?

Specially this new `AuthEvent` type.

jan-tennert:
> Hi @jan-tennert these changes deviates from the official libs, I'd like to better understand its benefits, could you explain?
> 
> Specially this new `AuthEvent` type.

@grdsdev 
Well as of right now, `auth-kt` doesn't really have an event system like the other client libs (which represent the `AuthEvent` as a String, what you wouldn't in Kotlin), `Auth#sessionStatus` is pretty much the replacement for that. The idea was to also check for error codes in deeplinks, urls, etc. aside from PKCE codes / access tokens and somehow make this receivable. Putting it in `sessionStatus` does not work because, this state should only represent session related data (Authenticated, NotAuthenticated, etc.), so we'd need something separate.
Other client libs have an event system for events like Sign Out, Sign In etc, but this already covered by `sessionStatus`, so I didn't add them and wanted to only add events which are not linked to a session state and can happen independently.
We could add them, but we'd have that overlap.

Issue #921, opened 2025-05-04
[Bug]: signUpWith(Email) returns null result


hieuwu:
### General Info

- [x] I checked for similar bug report
- [x] I am using the latest version
- [x] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.md) page for similar problems

### Version(s)

3.1.3

### Kotlin Target(s) and their respective versions

JVM 17, Kotlin 2.0.0, Android 14

### What happened? (include your code)

I am using  `signUpWith(Email)` to create new user. When the function is executed, it returns null while the user is still created on Supabase dashboard
Below is my code to create user 
```kotlin
val authResult = supabaseAuth.signUpWith(Email) {
    this.email = email
    this.password = password
    this.data = buildJsonObject {
        put("name", name)
    }
}
// authResult is null
```
User is created successfully on Dashboard

<img width="650" alt="Image" src="https://github.com/user-attachments/assets/82a1be6e-9f72-4ae8-84a4-55a28e0df5a0" />

### Steps To Reproduce (optional)

1. Use code from auth to create user 
```kotlin
val authResult = supabaseAuth.signUpWith(Email) {
    this.email = email
    this.password = password
}
// requireNotNul(authResult)
```
2. Assert `authResult` not null with `requireNotNul(authResult)`.
3. Expected result: The result should be available

PR #922, opened 2025-05-04
Fix null result when sign up with Email provider

hieuwu:
Decode result when result of `signUpWith(Email)` contains `access_token 

Result returns null
Related issue: https://github.com/supabase-community/supabase-kt/issues/921

## What is the new behavior?
Result should still be parsed to make it available instead of null

### DIFF
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/providers/builtin/DefaultAuthProvider.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/providers/builtin/DefaultAuthProvider.kt
@@ -19,6 +19,7 @@ import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.jsonObject
 
 /**
  * A default authentication provider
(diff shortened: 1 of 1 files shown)


jan-tennert (changes requested):
You probably have to adjust the tests, which are currently assuming that this can't happen

hieuwu:
> You probably have to adjust the tests, which are currently assuming that this can't happen

I have just updated. Please check again Jan

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/providers/builtin/DefaultAuthProvider.kt
```diff
@@ -95,10 +95,14 @@ sealed interface DefaultAuthProvider<C, R> : AuthProvider<C, R> {
...
-            onSuccess(userSession)
-            val userJson = json["user"]?.jsonObject ?: buildJsonObject { }
-            return decodeResult(userJson)
+            runCatching {
+                val userSession = supabaseJson.decodeFromJsonElement<UserSession>(json)
+                onSuccess(userSession)
+                val userJson = json["user"]?.jsonObject ?: buildJsonObject { }
+                return decodeResult(userJson)
+            }.onFailure {
+                return null
```
jan-tennert, line 104: I'd probably also add a `Auth.logger.w(exception) { "..." }` so the error doesn't just disappear

PR #927, opened 2025-05-11
Add new standard headers

### LINKED ISSUE #925: [Feature request]: Send standard client headers on all requests
### General Info

- [x] I installed the latest version of Supabase Kt
- [x] I checked for similar feature requests

### Feature request

For improving our internal usage reports, we need to send some client headers in all requests performed by the library.

Please include following headers:

- `x-sb-client-platform`: which platform of operating system on which the client is running `"Android"`, `"Chrome OS"`, `"Chromium OS"`, `"iOS"`, `"Linux"`, `"macOS"`, `"Windows"`
- `x-sb-client-platform-version`: version of the platform/operating system which the client is running `11.0.0` 
- `x-sb-client-model`: Indicates the device model on which the client is running, such as `Pixel 3` or `iPhone 15 Pro`, …

If value cannot be defined for some reason, send none.

Reference PR for Swift: https://github.com/supabase/supabase-swift/pull/685

### Usecase

_No response_

jan-tennert:

Feature (closes #925)

## What is the new behavior?

New standard headers providing OS name & version

### DIFF
--- a/Supabase/src/androidMain/kotlin/io/github/jan/supabase/PlatformTarget.android.kt
+++ b/Supabase/src/androidMain/kotlin/io/github/jan/supabase/PlatformTarget.android.kt
@@ -1,0 +1,6 @@
+package io.github.jan.supabase
+
+internal actual val OSInformation: OSInformation = OSInformation(
+    name = "Android",
+    version = android.os.Build.VERSION.SDK_INT.toString()
+)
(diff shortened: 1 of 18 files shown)


jan-tennert:
@grdsdev Should we distinguish between JS and WASM JS? Or combine them as "Browser"?

jan-tennert:
> > @grdsdev Should we distinguish between JS and WASM JS? Or combine them as "Browser"?
> 
> It is not right to consider WASM JS as browser, but I don't think there is an easy way of getting the platform running it, in this case is better to not send any value

Well both JS and WASM JS can run in the browser, but can also be server side, which we could check

PR #965, opened 2025-06-14
Introduce helpers to construct raw query string for FK data fetching with serializable class

hieuwu:
Introduce helpers to construct raw query string for FK data fetching with serializable class

Currently, when query data with foreign keys, we need to define a hardcoded string. For query that has one FK and multiple FK, they are different in format
- Strings are error-prone and hard to manage
- When `DTO` class changes, the related query has to be changed accordingly, now it is done manually

## What is the new behavior?
Introduce helpers to construct raw query string with serializable classes
- With new helpers, it work seamlessly with the mechanism of data class used to decode the response.
- Same `DTO class` can be used to construct query and decode data
- Raw query strings are updated accordingly when data class Json properties change
For example:
```kotlin
val query = RawQueryHelper.queryWithMultipleForeignKeys<MessageDto>()

// Returns
 """
content,
recipient: recipient(id, username, email),
sender: sender(id, username, email),
created_at
"""
 val columns = Columns.raw(query)
 val messages = postgrest.from("messages")
                .select(
                    columns = columns
                ) {
                    filter {
                         eq(queryParam, queryValue)
                     }
    }.decodeList<MessageDto>()

```


jan-tennert:
I know that specifying columns currently isn't ideal especially with FK, thats why I introduced #769, but to have somewhat of consistency between client libs I'm still waiting for implementations in other client libs

cc @grdsdev

hieuwu:
> I know that specifying columns currently isn't ideal especially with FKs, thats why I introduced #769, but to have somewhat of consistency between client libs I'm still waiting for implementations in other client libs. But great work, #769 has a more general approach for solving the columns problem via a KSP plugin
> 
> cc @grdsdev

Interesting. I also tried Ksp at the beginning but it took me too much time for general approach so I ended up just buildig helper functions 😂.
Hey Jan will there be any chance we can add KSP solution as experiment API? I am building project and use FK quite heavily, happy to test it.

jan-tennert:
> > I know that specifying columns currently isn't ideal especially with FKs, thats why I introduced #769, but to have somewhat of consistency between client libs I'm still waiting for implementations in other client libs. But great work, #769 has a more general approach for solving the columns problem via a KSP plugin
> > cc @grdsdev
> 
> Interesting. I also tried KSP at the beginning but it took me too much time for general approach so I ended up just build helper functions 😂. Hey Jan will there be any chance we can add KSP solution as experiment API? I am building project and use FK quite heavily, happy to test it.

Sure, I can publish a version. You can try `3.2.0-ksp-b1`, once [its done building](https://github.com/supabase-community/supabase-kt/actions/runs/15662689870)

Issue #1035, opened 2025-09-04
[Bug]: "." character in requests are not escaped


sproctor:
### General Info

- [x] I checked for similar bug report
- [x] I am using the latest version
- [x] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.md) page for similar problems

### Version(s)

3.2.2

### Kotlin Target(s) and their respective versions

JVM 21, Android

### What happened? (include your code)

I'm sending a string like "file.jpg" and the "." there messes up the request:

```
"GET | 200 | 24.35.100.105 | 979e932a2bb1394c | https://xxx.supabase.co/rest/v1/picture?and=%28account_id.eq.12345678-c6f3-4454-b9a6-55afd8fe49b5%2Cor%28modified_at.gt.2025-08-29T12%3A49%3A20.944065Z%2Cand%28modified_at.eq.2025-08-29T12%3A49%3A20.944065Z%2Ckey.gt.2Tare.jpg%29%29%29&order=modified_at.asc.nullslast%2Ckey.asc.nullslast&limit=1000&select=* | ktor-client"
```

Code:
```kotlin
supabase.from("public", "picture")
                .select {
                    filter {
                        if (since != null && lastId != null) {
                            and {
                                eq("account_id", accountId)
                                or {
                                    gt("modified_at", since)
                                    and {
                                        eq("modified_at", since)
                                        gt("key", lastId)
                                    }
                                }
                            }
                        } else {
                            eq("account_id", accountId)
                        }
                    }
                    order("modified_at", Order.ASCENDING)
                    order("key", Order.ASCENDING)
                    limit(limit.toLong())
                }
                .decodeList()
```

### Relevant log output (optional)

```shell

```

sproctor:
I'm not sure how to work around this one. Replacing the `.` with `%2E` results in `%252E` being sent.

PR #1036, opened 2025-09-05
make filter values comply with postrest rules

sproctor:

Bug fix for #1035

Values are passed through without being modified.

## What is the new behavior?

Values have backslashes and quotes added a backslash before, and values containing reserved characters are wrapped in quotes.

## Additional context

URL grammar is specified here: https://docs.postgrest.org/en/v13/references/api/url_grammar.html#reserved-chars

It's not clear if strings containing a backslash or quote need to be quoted.


jan-tennert:
Yea ignore that one, there are some issues with turbine. I'll look at the PR tomorrow.

jan-tennert (approved):

Issue #1154, opened 2026-01-27
[Bug]: Ktor 3.4.0 deprecates and breaks padding compatibility in `decodeBase64String`


sproctor:
I'll submit a PR shortly

PR #1155, opened 2026-01-27
Fix/base64

sproctor:

Fixes issue #1154. I'm not sure what happened with Ktor. This function used to handle both padded and unpadded values, but that changed in 3.4.0. They switched from a custom implementation to using `kotlin.io.encoding.Base64` with requiring padding. JWT's are URL-encoded without padding.


albertoalbaladejo:
Okay, thanks!

PR #1161, opened 2026-01-31
Fix default log level not correctly working

jan-tennert:

Bug fix, feature, docs update, ...

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context


jan-tennert:
> Does this work? It looks like the `SupabaseClient.LOGGER` always gets the initial state of the `DEFAULT_LOG_LEVEL`. Any changes need to be propagated the loggers. The only way I can see around this is to have a static default mutable config that is used when the min level isn't specified or read the `DEFAULT_LOG_LEVEL` on each log and either set the config or use `processLog` to have Kermit ignore the set min severity.

Yea realized that. Hmm we might need to refactor this completely, while doing that we can add an API for changing the logger

PR #1164, opened 2026-02-02
Partial ABI compatibility validation implementation

sproctor:

Very incomplete implementation of using Kotlin's ABI compatibility validation plugin. Before I went to implement this, I wanted to get your thoughts.

This could go in `buildSrc` and be applied to each module that should maintain ABI compatibility.

## Motivation

While contemplating some changes that would change the ABI, I thought it might be good to have a validation procedure in place.


jan-tennert:
Yea looks good to me at first glance. `@SupabaseInternal` marked methods/classes etc are still relevant for binary compatibility, no? I mean ignoring these would make sense but they still are relevant.

PR #1175, opened 2026-02-16
Allow nullable types in SupabaseSerializer and Postgrest decoding

ferhatwi:
Feature

The `SupabaseSerializer` interface and its related extension functions in `PostgrestResult` (like `decodeAs` and `decodeList`) enforce a `<T : Any>` type constraint.

The `<T : Any>` constraint creates a technical bottleneck where developers cannot natively decode to nullable types (e.g., `decodeAs<String?>()`). This forces a reliance on `decodeAsOrNull`, which conflates a valid database `NULL` with actual serialization errors, making it impossible to distinguish between missing data and malformed JSON without manual try-catch blocks.

## What is the new behavior?

This PR removes the `Any` constraint across the serialization layer, allowing `T` to be nullable.
- **Native Null Handling**: `decodeAs<T?>()` now correctly returns a Kotlin null when the server responds with a JSON null token.
- **List Support**: You can now decode lists containing null values (e.g., `decodeList<String?>()`) which was previously a compilation error.
- **Preserved Safety**: Strict null-safety is maintained. If a user calls `decodeAs<String>()` and the server returns `null`, the function will still throw a `SerializationException` as expected.
- **Moshi**: Now uses `serializeNulls`.

## Additional context

I have added 15 comprehensive unit tests to PostgrestTest to verify this behavior across several dimensions:
- Single Objects vs. Lists.
- Nullable vs. Non-nullable target types.
- Matching JSON & Result vs. Mismatching JSON & Result vs. JSON null.
- `decodeAs` (Strict) vs. `decodeAsOrNull` (Safe).


serializers/Jackson/src/commonTest/kotlin/JacksonSerializerTest.kt
```diff
@@ -13,13 +13,25 @@ class JacksonSerializerTest {
...
         assertEquals(value, decoded)
     }
 
+    @Test
+    fun testJacksonSerializerForNulls() {
+        val serializer = JacksonSerializer()
+        val supabaseClient = createMockedSupabaseClient(
+            configuration = {
+                defaultSerializer = serializer
+            })
```
jan-tennert:
```suggestion
            }
        )
```

serializers/Moshi/src/commonTest/kotlin/MoshiSerializerTest.kt
```diff
@@ -13,13 +13,25 @@ class MoshiSerializerTest {
         val supabaseClient = createMockedSupabaseClient(
             configuration = {
                 defaultSerializer = serializer
-            }
-        )
+            })
```
jan-tennert:
```suggestion
            }
        )
```

PR #1254, opened 2026-04-16
Use inject-based logging approach for configurabilty and mocking

### LINKED ISSUE #1247: [Feature request]: A way to install completely custom logging
### General Info

- [x] I installed the latest version of supabase-kt
- [x] I checked for similar feature requests

### Feature request

I would love to be able to custom handle my Supabase logging messages

### Usecase

I want to send certain messages to my Sentry logs. Currently they printed with no way to intercept them on JVM or sent to logcat and I don't want to send all of my logcat stuff to sentry.

jan-tennert:

Rework (closes #1247)

Loggers are static, which has the upside that they can be called from anywhere, but customization is hard and messy.

## What is the new behavior?

In the `SupabaseClientBuilder`, you specify two things:
- `defaultLogLevel` - Basically the same as before, this is the log level for all loggers, if no other is specified
- `defaultLoggingFactory` - A factory for creating a `SupabaseLoggingProcessor` e.g. the default:
```kotlin
defaultLoggingFactory = { level -> KermitLoggingProcessor(level) }
```

Then a `SupabaseLogger` gets created in the `SupabaseClient` using the two options. This class basically just calls the processor, but has some handy methods on top.

On the plugin side you can still override the log level and processor, if you want for some reason. Each plugin has a `logger` property now.

Todo:
- [ ] Tests

### DIFF
--- a/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
+++ b/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
@@ -37,17 +37,17 @@ private fun addLifecycleCallbacks(auth: Auth) {
 
                 override fun onStart(owner: LifecycleOwner) {
                     if(!auth.isAutoRefreshRunning && auth.config.alwaysAutoRefresh) {
-                        Auth.logger.d {
+                        auth.logger.d {
                             "Trying to re-load session from storage..."
                         }
                         scope.launch {
                             val sessionFound = auth.loadFromStorage()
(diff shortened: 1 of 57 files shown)


huudev:
Hey bro, I might be mistaken, but I think there could be a small issue with
fun appendTag(tag: String) = withTag(tag + this.tag)

Maybe it should be:
fun appendTag(tag: String) = withTag(this.tag + tag)?

jan-tennert:
Yea thats backwards, opened a PR.

PR #1267, opened 2026-04-26
fix(auth): don't emit spurious NotAuthenticated on Android lifecycle onStart

### LINKED ISSUE #1266: auth-kt-android — ProcessLifecycleOwner.onStart race emits spurious NotAuthenticated(isSignOut=false) while session refresh is in flight
### Version

`io.github.jan-tennert.supabase:auth-kt:3.5.0` (`auth-kt-android`)

### Description

When the app returns to the foreground with an expired session in storage, `addLifecycleCallbacks`'s `onStart` callback briefly emits `SessionStatus.NotAuthenticated(isSignOut = false)` _while_ the async refresh job it just launched is still in flight. A few hundred milliseconds later the refresh succeeds and `Authenticated` is emitted. Consumers observing `sessionStatus` see a transient sign-out flash even though nothing was actually signed out.

In our app this surfaced as the auth/welcome screen briefly animating in for ~1s on every background→foreground transition that exceeded the access-token lifetime.

### Repro

1. Sign in.
2. Background the app long enough for the access token to fall under `SESSION_REFRESH_THRESHOLD` (i.e. an expired session is what `loadFromStorage` will find).
3. Foreground the app.
4. Observe `auth.sessionStatus` — you'll see `Authenticated → Initializing (onStop) → NotAuthenticated(isSignOut=false) → Authenticated`.

### Expected

`onStart`'s foreground re-entry should not emit `NotAuthenticated` when there *is* a session in storage and a refresh is being attempted. The terminal state should be `Authenticated` (refresh succeeds), `RefreshFailure` (transient/5xx), or `NotAuthenticated(isSignOut=true)` (refresh 4xx → `clearSession`). Never `NotAuthenticated(isSignOut=false)` purely because of timing.

### Root cause

[`Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt`](https://github.com/supabase-community/supabase-kt/blob/master/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt), `addLifecycleCallbacks`:

```kotlin
override fun onStart(owner: LifecycleOwner) {
    if (!auth.isAutoRefreshRunning && auth.config.alwaysAutoRefresh) {
        scope.launch {
            val sessionFound = auth.loadFromStorage()  // initializing=false (default)
            // … logging …
            auth.initDone()                             // ← runs immediately
        }
    }
}
```

`auth.loadFromStorage()` resolves to the public override, which calls the internal `loadFromStorage(autoRefresh, initializing = false)` in `AuthImpl`. For an expired session that path calls `recreateSessionJob(session, source, delay = false, initializing = false)`, and `recreateSessionJob` only `sessionJob?.join()`s when `initializing == true`:

```kotlin
private suspend fun recreateSessionJob(
    session: UserSession, source: SessionSource, delay: Boolean, initializing: Boolean
) {
    sessionJob?.cancel()
    sessionJob = authScope.launch { /* refresh */ }
    if (initializing) sessionJob?.join()
}
```

So control returns to the lifecycle coroutine before the refresh has had a chance to set the status to `Authenticated`. The next line (`auth.initDone()`) sees `_sessionStatus == Initializing` (set by `onStop`) and emits `NotAuthenticated(isSignOut = false)`.

`Auth.init()`'s cold-start path doesn't have this bug because it goes through `loadFromStorage(initializing = true)`, which *does* join.

### Suggested fix

Either:

**A.** Don't call `initDone()` when a session was loaded — `importSession` will eventually set the status, and on refresh failure `clearSession()` / `RefreshFailure` will:

```kotlin
scope.launch {
    val sessionFound = auth.loadFromStorage()
    if (!sessionFound) auth.initDone()  // only when storage was genuinely empty
}
```

**B.** Make the lifecycle path use the `initializing = true` overload so the refresh is joined before `initDone()` runs (mirrors the cold-start `Auth.init()` path).

(A) is simpler and matches what consumers typically expect. (B) preserves the "always end Initializing" guarantee but blocks `onStart` on network — probably worse for foreground latency.

### Workaround

In our `AuthRepository` we now treat `NotAuthenticated(isSignOut = false)` as recoverable when our UI state is already signed-in, routing through `auth.loadFromStorage()` instead of flipping straight to a signed-out UI. This avoids the visible flash but masks the underlying race.

### Environment

- Pixel 9 Pro, Android 16
- Kotlin Multiplatform shared module (Compose Multiplatform + Voyager)
- supabase-kt 3.5.0

alexeysfh:
## Summary

Fixes #1266 — on background→foreground with an expired session in storage, the Android lifecycle observer emits a transient `NotAuthenticated(isSignOut=false)` while the async refresh is still in flight. Consumers see a spurious sign-out flash before `Authenticated` arrives.

## Root cause (full analysis in #1266)

`addLifecycleCallbacks.onStart` calls `loadFromStorage()` then `initDone()` in the same coroutine. For an expired session, `loadFromStorage()` queues `recreateSessionJob` in `authScope.launch{}` and **does not join** because `initializing=false`. Control returns to the lifecycle coroutine which immediately calls `initDone()` while `_sessionStatus` is still `Initializing`. `initDone()` then emits `NotAuthenticated(isSignOut=false)` even though a valid session was loaded and a refresh is in progress.

`Auth.init()`'s cold-start path doesn't have this bug because it uses `loadFromStorage(initializing=true)` which joins the refresh job.

## Change

Only call `initDone()` when no session was found. When a session is loaded, `importSession()` sets the terminal status itself — `Authenticated` on refresh success, `NotAuthenticated(isSignOut=true)` via `clearSession()` on 4xx, or `RefreshFailure` on transient errors.

## Test

The existing `PlatformSetupTest` runs with `enableLifecycleCallbacks = false`, so the buggy path isn't exercised. I haven't added a lifecycle-driven test in this PR — happy to add one (Robolectric or similar to drive `ProcessLifecycleOwner`) if you'd prefer that before merging.

### DIFF
--- a/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
+++ b/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
@@ -51,15 +51,6 @@ private fun addLifecycleCallbacks(auth: Auth) {
                                 auth.logger.d {
                                     "Session found, auto refresh started"
                                 }
-                                // Do NOT call initDone() here. importSession()
-                                // launched by loadFromStorage() will set the
-                                // status to Authenticated once the (possibly
-                                // async) refresh completes — or to
-                                // NotAuthenticated via clearSession() on a 4xx
-                                // refresh failure. Calling initDone() while the
-                                // refresh job is still in flight emits a spurious
-                                // NotAuthenticated(isSignOut=false) because the
-                                // status is still Initializing at that point.
                             }
                         }
                     }
--- a/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
+++ b/Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
@@ -46,12 +46,21 @@ private fun addLifecycleCallbacks(auth: Auth) {
                                 auth.logger.d {
                                     "No session found, not starting auto refresh"
                                 }
+                                auth.initDone()
                             } else {
                                 auth.logger.d {
                                     "Session found, auto refresh started"
                                 }
+                                // Do NOT call initDone() here. importSession()
+                                // launched by loadFromStorage() will set the
+                                // status to Authenticated once the (possibly
+                                // async) refresh completes — or to
+                                // NotAuthenticated via clearSession() on a 4xx
+                                // refresh failure. Calling initDone() while the
+                                // refresh job is still in flight emits a spurious
+                                // NotAuthenticated(isSignOut=false) because the
+                                // status is still Initializing at that point.
                             }
-                            auth.initDone()
                         }
                     }
                 }


Auth/src/androidMain/kotlin/io/github/jan/supabase/auth/setupPlatform.kt
```diff
@@ -46,12 +46,21 @@ private fun addLifecycleCallbacks(auth: Auth) {
...
                                 }
+                                // Do NOT call initDone() here. importSession()
+                                // launched by loadFromStorage() will set the
+                                // status to Authenticated once the (possibly
+                                // async) refresh completes — or to
+                                // NotAuthenticated via clearSession() on a 4xx
+                                // refresh failure. Calling initDone() while the
+                                // refresh job is still in flight emits a spurious
+                                // NotAuthenticated(isSignOut=false) because the
+                                // status is still Initializing at that point.
```
jan-tennert:
```suggestion
```

I think that should be clear enough without docs

jan-tennert (approved):

PR #1275, opened 2026-04-28
Improve test coverage in Postgrest

hieuwu:
Improve test coverage of Postgrest

There are missing tests for behavior of requests and classes in Postgrest module

## What is the new behavior?
Cover tests for missing functionalities

## Additional context
Before
Screenshot 2026-04-28 at 16 22 02

After
Screenshot 2026-04-28 at 16 34 58


Postgrest/src/commonTest/kotlin/PostgrestUpdateTest.kt
```diff
@@ -0,0 +1,298 @@
...
+        val testGeneric: TestInnerObj
+    )
+
+    @Serializable
+    data class TestInnerObj(val innerMsg: String)
+
+    private val serializer = KotlinXSerializer()
+
+    @Test
+    fun testSetToGeneric() {
```
jan-tennert: These are probably failing because they are only available on JVM/Android

hieuwu: Let me see

hieuwu: I change the default value of `buildPostgrestUpdate` to be `CAMEL_CASE_TO_SNAKE_CASE` at https://github.com/supabase-community/supabase-kt/pull/1275/commits/3413b966fcb789b0b29660591ec28fb678969a7c
Default value of `propertyConversionMethod` in the  `Config` also the same so it should not lead to any breaking change

Postgrest/src/commonTest/kotlin/PostgrestUpdateTest.kt
```diff
@@ -0,0 +1,298 @@
...
+import io.github.jan.supabase.serializer.KotlinXSerializer
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class PostgrestUpdateTest {
```
jan-tennert: Do we need that many tests for null values?

Postgrest/src/commonTest/kotlin/PostgrestTest.kt
```diff
@@ -130,6 +184,49 @@ class PostgrestTest {
...
+                assertEquals(expectedBody, receivedBody)
+                assertEquals("data", it.url.parameters["columns"])
+                assertEquals("schema", it.headers["Content-Profile"])
+                respond("")
+            }
+        )
+    }
+
     @Test
     fun testUpsert() {
```
jan-tennert: I'm not sure if that test is still necessary if we have those other two

Postgrest/src/commonTest/kotlin/PostgrestTest.kt
```diff
@@ -105,6 +142,23 @@ class PostgrestTest {
...
+            requestHandler = {
+                assertMethodIs(HttpMethod.Head, it.method)
+                assertEquals(columns.value, it.url.parameters["select"])
+                respond("")
+            }
+        )
+    }
+
     @Test
     fun testInsert() {
```
jan-tennert: same with this one

Postgrest/src/commonTest/kotlin/PostgrestTest.kt
```diff
@@ -155,6 +252,49 @@ class PostgrestTest {
...
+                assertEquals(expectedBody, receivedBody)
+                assertEquals("data", it.url.parameters["columns"])
+                assertEquals("schema", it.headers["Content-Profile"])
+                respond("")
+            }
+        )
+    }
+
     @Test
     fun testUpdate() {
```
jan-tennert, line 207: probably a few of the tests in here might be duplicates

Postgrest/src/commonTest/kotlin/PostgrestUpdateTest.kt
hieuwu: It is to cover all supported types in the request. I removed the test file at https://github.com/supabase-community/supabase-kt/pull/1275/changes/01c2b78625214225a31d9f61797a0d6f6e56862f. Seems it makes no sense just to test get, set function. The increased coverage is not worth though.

Postgrest/src/commonTest/kotlin/PostgrestTest.kt
hieuwu: @jan-tennert I simplified the test at https://github.com/supabase-community/supabase-kt/pull/1275/commits/5d129ecc42bbc720648128d6127a821697223266. 

This is the latest coverage. > 80% is fine now.
Screenshot 2026-05-07 at 21 20 42

hieuwu: Also I've noticed that overall currently there are many ways to do the same thing like get, set and build object, which makes many public function calls to the same function. That's why when try to increase coverage for just small number of lines it leads to lots of duplicate tests. 
I'll try to analyze and find some way to improve the test structure

jan-tennert (approved):

PR #1283, opened 2026-05-07
OAuth / deeplink refactor for Android and iOS

jan-tennert:
## What kind of change does this PR introduce?

Feature (closes #1253)

**Android platform improvements:**

* Added manifest placeholders (`supabase.auth.scheme` and `supabase.auth.host`) and an intent-filter for `SupabaseAuthDispatcherActivity` in `AndroidManifest.xml` to enable automatic deep link handling for authentication callbacks.
* Introduced `SupabaseAuthDispatcherActivity` to handle external authentication flows and redirects, centralizing the logic for launching browsers/custom tabs and processing redirect URIs.
* Replaced the previous `openUrl` logic with intent-based dispatching to the new activity, and updated `openExternalUrl` to use this mechanism.
* Added `AuthFlowManager` for managing redirect URIs via a shared coroutine flow, enabling asynchronous deep link handling across the app lifecycle.
* Updated platform setup to listen for deep links and process them using the new flow manager.

**Apple platform improvements:**

* Introduced a similar `AuthFlowManager` for iOS/macOS, using a shared coroutine flow to handle incoming authentication redirects.
* Refactored deep link handling to use the flow manager, and updated setup to automatically listen for and process authentication deep links.

**Common and configuration changes:**

* Deprecated manual setting of `scheme` and `host` in favor of manifest-based configuration on Android and `appScheme` on iOS.
* Moved and refactored `ExternalAuthAction` into its own file, making it `Parcelable` for intent passing.
* Updated `defaultPlatformRedirectUrl` on Android to read redirect info from manifest metadata.

# Example

**Android**

On Android, the only thing you have to do now is set the app scheme and optionally the host in the build configuration:
```kotlin
android {

    defaultConfig {
        manifestPlaceholders["supabaseAuthScheme"] = "myScheme"
        // optional: 
        manifestPlaceholders["supabaseAuthHost"] = "myHost"
    }
    
}
```

The library will take care of everything else.

**iOS**

On iOS, `ASWebAuthenticationSession` is now used for OAuth, fixing #1253 and also making the process more iOS idiomatic. For magic links / deeplinks from outside the app, you still have to use `handleDeeplinks()`, but with a new method for better visbility in Swift (`SupabaseDeeplinkHandlerKt.handleDeeplinks()`. The scheme entry in `PList.info` is still required as is setting the scheme in the config:

```
install(Auth) {
     appScheme = "myScheme" 
}
```

Issue #1253, opened 2026-04-16
[Question]: Apple reject my app for external browser login


abalta:
### General info

- [x] I checked the issues and discussions for similar questions
- [x] I checked the [troubleshooting](https://github.com/supabase-community/supabase-kt/blob/master/TROUBLESHOOTING.md) page for similar problems

### What is your question?

Hi, I have Compose Multiplatform application. My application already published Google Play/App Store. I use Apple Login (sign-in natively with in app browser), Twitch and Discord login for external browser.

Apple reviewer rejected for Twitch and Discord login. Does native sign in support Twitch and Discord provider?

Here is reviewer message:

Bug Fix Submissions

The issues we've identified below are eligible to be resolved on your next update. If this submission includes bug fixes and you'd like to have it approved at this time, reply to this message and let us know. You do not need to resubmit your app for us to proceed.

Alternatively, if you'd like to resolve these issues now, please review the details, make the appropriate changes, and resubmit.

Review Environment
Submission ID: 3fbc688a-2782-469c-9b3b-2408ccb9cf31
Review date: April 16, 2026
Review Device: iPhone 17 Pro Max and iPad Air 11-inch (M3) 
Version reviewed: 1.9

Guideline 4 - Design

Issue Description

We noticed that the user is taken to the default web browser to sign in or register for an account, which provides a poor user experience.

Next Steps

To resolve this issue, please revise the app to enable users to sign in or register for an account in the app.

You may also choose to implement the Safari View Controller API to display web content within the app. The Safari View Controller allows the display of a URL and inspection of the certificate from an embedded browser in an app so that customers can verify the webpage URL and SSL certificate to confirm they are entering their sign in credentials into a legitimate page.

### Relevant log output (optional)

```shell

```

jan-tennert:
@abalta Sorry for the late reply

> Apple reviewer rejected for Twitch and Discord login. Does native sign in support Twitch and Discord provider?

No there is no Native Sign in for Twitch or Discord.

I'm not sure if it possible to implement OAuth using the Safari View Controller API within the library, that might be something to implement yourself. If there is a good way to implement this within a library then sure we can look into it, but the library can't handle everything. Related https://github.com/supabase-community/supabase-kt/issues/1094

abalta:
Thanks for reply @jan-tennert, I'll check that later on.

PR #1291, opened 2026-05-14
Simplified OAuth, SSO, Native Auth in separate module

jan-tennert:

Feature, the target branch is probably going to change, since this PR would be included in 4.0

`auth-kt` does two things: mirroring the JS SDK in terms of API methods but also providing platform specific implementations for stuff like OAuth, Deeplinks, etc. This approach worked, but is now getting even more bloated with the addition of passkeys and a proposed API in #1283 (which is now included in this PR). Also, people not using these platform-specific implementation, still have the needed dependencies included, which is not ideal.

## What is the new behavior?

Everything below is work-in-progress. Note that if this PR goes through with the proposed changes, `compose-auth` might get obsolete and removed in 4.0

### Base changes

**`auth-kt` really only provides:**
(1) Methods to call the Supabase API (now with split `auth.signInWithPassword()`, `auth.signInWithOtp()`, `auth.signInWithIdToken()`, etc for SDK parity)
(2) Handle session management, if not disabled (this PR also makes all API related methods return a `UserSession` and also an option to disable the session importing into memory, so it is even easier for server-side applications)

**`auth-native` provides:**
- Seamless platform specific implementations for OAuth, Native Auth, Passkeys, Lifecycle Management, Deeplinks (of depending on the platform)

The only thing you have to do:
```kotlin
install(Auth.withNative())
```

Note that the base `auth-kt` doesn't have any of these methods, since they would need platform specific dependencies. If users decide to make their own implementation they can just use `auth.getOAuthUrl(...)` and `auth.passkeys`

### New syntax

**Sign in [With Email() or Phone()]**
```kotlin
val session = supabase.auth.signInWithPassword(Email("myEmail@example.com"), password = "123345") {
        // ...
}
```

**Sign in with OTP**
```kotlin
supabase.auth.signInWithOtp(Email("myEmail@example.com")) {
      // ...
}
```

**Sign up [with Email() or Phone()]**
```kotlin
val response = supabase.auth.signUp(Email("myEmail@example.com"), password = "12345") {
    // ...
}
println(response.session)
println(response.user)
```

**Sign in with ID Token**
```kotlin
val session = supabase.auth.signInWithIdToken(OAuthProviders.GOOGLE, token = "12343") {
    // ...
}
```
__**Native-Auth module:**__
**Sign in with OAuth**
```kotlin
supabase.auth.signInWithOAuth(OAuthProviders.GOOGLE) {
    // ...
}
```

**Linking identity**
```kotlin
supabase.auth.linkIdentity(OAuthProviders.GOOGLE) {
    // ...
}
```

**Sign in with SSO**
```kotlin
supabase.auth.signInWithSSO(SSODomain("domain")) 
supabase.auth.signInWithSSO(SSOProvider("providerId"))
```

**Native Google Auth (or OAuth as a fallback)** 
```kotlin
// required for Android:
install(Auth.withNative()) {
    nativeAuth {
         googleClientId = "..."
    }
}
val result = supabase.auth.signWithGoogle()
when(result) {
    GoogleSignInResult.Cancelled -> TODO()
    GoogleSignInResult.OAuthInitiated -> TODO()
    is GoogleSignInResult.Success -> TODO()
}
```

**Native Apple Auth (or OAuth as a fallback)**
```kotlin
val result = supabase.auth.signWithApple()
when(result) {
    AppleSignInResult.Cancelled -> TODO()
    AppleSignInResult.OAuthInitiated -> TODO()
    is AppleSignInResult.Success -> TODO()
}
```

### Android / iOS OAuth related changes

**Android**

On Android, the only thing you have to do now is set the app scheme and optionally the host in the build configuration:
```kotlin
android {

    defaultConfig {
        manifestPlaceholders["supabaseAuthScheme"] = "myScheme"
        // optional: 
        manifestPlaceholders["supabaseAuthHost"] = "myHost"
    }
    
}
```

The library will take care of everything else. No more `handleDeeplinks()` or something similar. The library will also catch magic links.

**iOS**

On iOS, `ASWebAuthenticationSession` is now used for OAuth, fixing #1253 and also making the process more iOS idiomatic. For magic links / deeplinks from outside the app, you still have to use `handleDeeplinks()`, but with a new method for better visbility in Swift (`SupabaseDeeplinkHandlerKt.handleDeeplinks()`. The scheme entry in `PList.info` is still required as is setting the scheme in the config:

```
install(Auth) {
     appScheme = "myScheme" 
}
```

Todos:
- [ ] Fix tests
- [x] Think about maybe making it possible to use the credentials while signing in (so you can set them in the user's data)
- [ ] KDocs
- [ ] Documentation
- [ ] Passkeys
- [x] ~Google Auth on iOS~ Will probably not happen in this PR, waiting for SwiftPM getting more stable


hieuwu:
I wonder now that we have ` auth-native` , how about ones in `supabase-kt-plugins` ? Should native auth related code in plugins  be moved to `auth-native` ?

jan-tennert:
> I wonder now that we have ` auth-native` , how about ones in `supabase-kt-plugins` ? Should native auth related code in plugins be moved to `auth-native` ?

Do you mean what happens to compose auth?
> Everything below is work-in-progress. Note that if this PR goes through with the proposed changes, compose-auth might get obsolete and removed in 4.0

hieuwu:
> > I wonder now that we have ` auth-native` , how about ones in `supabase-kt-plugins` ? Should native auth related code in plugins be moved to `auth-native` ?
> 
> Do you mean what happens to compose auth?
> 
> > Everything below is work-in-progress. Note that if this PR goes through with the proposed changes, compose-auth might get obsolete and removed in 4.0

Interesting. This would make the integration more convinient 👍

jan-tennert:
cc @grdsdev

AuthNative/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+                withMingw()
+                withJvm()
+                withWasmJs()
+                withJs()
+                withMacos()
+            }
+        }
+    }
+    allTargets()
+   /* swiftPMDependencies {
```
hieuwu, line 54: Nice one @jan-tennert  👍. I was about to ask if we can replace `spm4kmp` from the plugins to this. Let me know if you need some iOS testing

jan-tennert: Yea I'll probably leave that one out of this PR. Also not completely sure how that swift pm integration works with publishing yet

PR #1300, opened 2026-06-06
implement structured X-Client-Info header

### LINKED ISSUE #1297: feat: implement structured X-Client-Info header
## Background

Supabase is consolidating client metadata into a single structured `X-Client-Info` header instead of sending multiple separate headers. The motivation is that adding a new HTTP header in supabase-js is a breaking change for applications running in Edge Functions — users must explicitly allowlist headers in their CORS config. Since `X-Client-Info` is already in every allowlist, extending it as a structured value lets all SDKs add new metadata fields freely.

## New format

```
X-Client-Info: name/version[; key=value ...]
```

Parameters are semicolon-delimited `key=value` pairs after the library token (same convention as `Content-Type`). Order is not guaranteed — parsers must locate by key.

**Example:**
```
X-Client-Info: supabase-kt/2.0.0; platform=Android; platform-version=14.0; runtime=kotlin; runtime-version=1.9.0
```

## Parameters to include

| Parameter | Description | Notes |
| --- | --- | --- |
| `platform` | OS / host environment | `Android`, `iOS`, `macOS`, `Windows`, `Linux` (via KMP targets) |
| `platform-version` | OS version string | Follow host platform convention |
| `runtime` | Language runtime | `kotlin` |
| `runtime-version` | Runtime version | e.g. `1.9.0` |
| `framework` | App framework if detectable | `compose`, `ktor`, etc. |
| `framework-version` | Framework version | |

All parameters are optional — omit entirely when the value cannot be determined. Do not use empty strings or `unknown` as values.

## Spec

Full behavior contract: https://github.com/supabase/sdk/pull/3

## Deprecated headers to stop sending

Once this is implemented, the following separate headers should no longer be sent:
- `X-Supabase-Client-Platform`
- `X-Supabase-Client-Platform-Version`
- `X-Supabase-Client-Runtime`
- `X-Supabase-Client-Runtime-Version`

(Remove them only if this SDK currently sends them — no action needed if they were never added.)

jan-tennert:

Feature (closes #1297)

Old header structure

## What is the new behavior?

New header structure as described in #1297 

## Additional context

`framework`, `framework-version` is not detectable.

### DIFF
--- a/Supabase/src/commonMain/kotlin/io/github/jan/supabase/network/KtorSupabaseHttpClient.kt
+++ b/Supabase/src/commonMain/kotlin/io/github/jan/supabase/network/KtorSupabaseHttpClient.kt
@@ -126,6 +126,7 @@ class KtorSupabaseHttpClient @SupabaseInternal constructor(
                 }
             }
             put("runtime", "kotlin")
+            put("runtime-version", KotlinVersion.CURRENT.toString())
         }
         append("X-Client-Info", "supabase-kt/${BuildConfig.PROJECT_VERSION}" + metadata.toList().joinToString("; ", prefix = "; ") { "${it.first}=${it.second}" })
     }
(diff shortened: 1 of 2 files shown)


grdsdev (commented):
There's not way to know the Kotlin version being used to compile the package?

jan-tennert:
> There's not way to know the Kotlin version being used to compile the package?

Actually, it's possible to get the current stdlib version so we can use that

PR #1316, opened 2026-06-09
Native Email Verification (Android only)

hieuwu:
Example to show case native email verification https://github.com/supabase-community/supabase-kt-plugins/pull/96

Currently user need to open email, click link and when open the app via app link, the app need to perform email verification. 
Confirmation email may go to Trash mailbox and hard for them to verify email

## What is the new behavior?
Verify their email directly

## Additional context
Supabase API has not supported email verification with sd-jwt yet. We need to use Edge Function or backend to perform email confirmation status when user try to verify their email

https://github.com/user-attachments/assets/dab8dd21-e712-48f3-9c5a-4e90e499e400


sample/chat-demo-mpp/common/build.gradle.kts
```diff
@@ -52,7 +52,10 @@ kotlin {
                 api(compose.foundation)
                 api(compose.material3)
                 api(compose.materialIconsExtended)
-                addModules(SupabaseModule.AUTH, SupabaseModule.POSTGREST, SupabaseModule.REALTIME)
+                addModules(SupabaseModule.AUTH, SupabaseModule.POSTGREST, SupabaseModule.REALTIME, SupabaseModule.FUNCTIONS)
+                api("io.github.jan-tennert.supabase:compose-auth:$pluginVersion")
```
hieuwu, line 56: Just to test with plugins, may remove later

PR #1319, opened 2026-06-14
fix: use `loadSessionOrNull()` for `loadFromStorage()`

### LINKED ISSUE #1318: auth: loadFromStorage logs "Failed to load session" at ERROR for the normal no-session case (e.g. cold start while logged out)
### Version

3.6.0 (also present on `master` as of June 2026)

### Summary

`AuthImpl.loadFromStorage()` logs `Failed to load session` at **ERROR with a full stacktrace** whenever there is no stored session — which is a normal, expected state (any logged-out user). It is caught (no crash, `loadFromStorage` returns `false`), but the ERROR-level log + stacktrace is misleading: it looks like something failed when nothing did.

This is triggered on a **plain cold start while logged out** via the `autoLoadFromStorage` path (commonMain → affects **all platforms**), and additionally on Android via the `onStart` lifecycle callback after `signOut()`.

### Stacktraces

Cold start, no stored session (`autoLoadFromStorage`, `AuthImpl.init` — commonMain, all platforms):

```
Failed to load session
java.lang.IllegalStateException: No entry with the key sb-<project-ref>-supabase-co-session
	at io.github.jan.supabase.auth.SettingsSessionManager.loadSession(SettingsSessionManager.kt:51)
	at io.github.jan.supabase.auth.AuthImpl.loadFromStorage(AuthImpl.kt:685)
	at io.github.jan.supabase.auth.AuthImpl.loadFromStorage$default(AuthImpl.kt:684)
	at io.github.jan.supabase.auth.AuthImpl$init$2.invokeSuspend(AuthImpl.kt:149)
```

Android, after `signOut()` + foreground (`addLifecycleCallbacks` `onStart`):

```
Failed to load session
java.lang.IllegalStateException: No entry with the key sb-<project-ref>-supabase-co-session
	at io.github.jan.supabase.auth.SettingsSessionManager.loadSession(SettingsSessionManager.kt:51)
	at io.github.jan.supabase.auth.AuthImpl.loadFromStorage(AuthImpl.kt:685)
	at io.github.jan.supabase.auth.SetupPlatformKt$addLifecycleCallbacks$1$1$onStart$2.invokeSuspend(setupPlatform.kt:44)
```

### Root cause

- `SettingsSessionManager.loadSession()` does `getStringOrNull(key) ?: error("No entry with the key $key")`, i.e. it throws when there is simply no stored session.
- `AuthImpl.loadFromStorage()` does roughly:
  ```kotlin
  val session = try { sessionManager.loadSession() } catch (e: Exception) {
      logger.e(e) { "Failed to load session" }
      null
  }
  ```
  so it logs **any** exception at ERROR, including the expected empty-session case.
- `AuthImpl.init` calls `loadFromStorage(initializing = true)` when `autoLoadFromStorage` is true (the default), so every launch without a stored session logs this on every platform.
- On Android, `addLifecycleCallbacks` `onStart` also calls `loadFromStorage()` when auto-refresh isn't running and `alwaysAutoRefresh` is true (e.g. after `signOut()` followed by a foreground transition).

### Expected behavior

The "no stored session" case shouldn't be logged at ERROR with a stacktrace. A couple of options:

- Use the already-existing `loadSessionOrNull()` in `loadFromStorage` instead of `loadSession()` + try/catch; or
- Distinguish "no stored session" (expected → debug log) from "a session existed but failed to deserialize" (real error → error log).

This is adjacent to #1314 (returning null from session storage on JSON parse failure).

### To Reproduce

1. Launch the app with no stored session (fresh install, or after `signOut()`).
2. Observe the `Failed to load session` ERROR log with stacktrace on startup.

(On Android it also reproduces by sending the app to background and back while logged out.)

jan-tennert:

closes #1318, #1314

`loadFromStorage()` uses `loadSession()` and treats the exception thrown when no session is found as a critical exception

## What is the new behavior?

The method now correctly uses the nullable method

## Additional context

### DIFF
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthImpl.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AuthImpl.kt
@@ -682,7 +682,7 @@ internal class AuthImpl(
     override suspend fun loadFromStorage(autoRefresh: Boolean): Boolean = loadFromStorage(autoRefresh, false)
 
     suspend fun loadFromStorage(autoRefresh: Boolean = config.alwaysAutoRefresh, initializing: Boolean): Boolean {
-        val session = try { sessionManager.loadSession() } catch (e: Exception) {
+        val session = try { sessionManager.loadSessionOrNull() } catch (e: Exception) {
             currentCoroutineContext().ensureActive()
             logger.e(e) { "Failed to load session" }
             null
--- a/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/SessionManager.kt
+++ b/Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/SessionManager.kt
@@ -21,11 +21,7 @@ interface SessionManager {
     /**
      * Loads the saved session from storage.
      */
-    suspend fun loadSessionOrNull(): UserSession? = try {
-        loadSession()
-    } catch(e: Exception) {
-        null
-    }
+    suspend fun loadSessionOrNull(): UserSession?
 
     /**
      * Deletes the saved session from storage.
@@ -46,11 +42,15 @@ class MemorySessionManager(session: UserSession? = null): SessionManager {
     }
 
     override suspend fun loadSession(): UserSession {
-        return session.load() ?: error("No session stored")
+        return loadSessionOrNull() ?: error("No session stored")
     }
 
     override suspend fun deleteSession() {
         session.store(null)
+    }
+
+    override suspend fun loadSessionOrNull(): UserSession? {
+        return session.load()
     }
 
 }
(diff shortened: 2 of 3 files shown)


PierreVieira:
For your reference: this one can be better https://github.com/supabase-community/supabase-kt/pull/1320, it adds some tests and make an explicit exception that'll be ignored for this case.
You choose anyway, thanks!

jan-tennert:
@PierreVieira, yea this one wasn't ready. We can merge both into yours.

PierreVieira:
Sounds good, thanks! I've updated #1320 to also close #1314:

- "no session" → debug log, "corrupt session" → still an ERROR log (real problem), both treated as not-logged-in without crashing.
- `SessionManager.loadSessionOrNull()` keeps its default impl, so it stays source-compatible for custom `SessionManager` implementations.
- Added a test for the corrupt-session case (`loadSessionOrNull()` returns `null`).

One open decision noted in the PR: for #1314 I kept the corrupt-session case logged at ERROR for observability rather than silencing it entirely like supabase-js does — happy to switch to a silent `null` if you'd prefer strict parity.

PR #1320, opened 2026-06-14
fix(auth): don't log an error when there is no stored session

### LINKED ISSUE #1314: parity(auth): return null from session storage on JSON parse failure [from supabase-js]
> [!WARNING]
> **Auto-generated parity issue — may be a false positive.**
>
> This issue was created automatically by `/sync-sdk-parity` from a heuristic analysis of recent `supabase-js` commits. The tooling has limited insight into language-specific idioms and may have:
>
> - misidentified a JS-only change as cross-language relevant,
> - missed an existing implementation in this SDK under a different name,
> - or proposed an API shape that doesn't fit this language's conventions.
>
> **It is the SDK author's responsibility to validate the need before implementing.** If this change does not apply to this SDK, please close the issue with a short note explaining why.

---

## SDK Parity: Kotlin implementation needed

A defensive fix was made in `supabase-js` auth that may apply to this repository. Please confirm applicability before starting work.

## Reference Implementation (supabase-js)

- **Commit**: [`0c227a1`](https://github.com/supabase/supabase-js/commit/0c227a1)
- **PR**: supabase/supabase-js#2336
- **Module**: auth
- **Type**: bug-fix (resilience)

### What Changed

The internal session-storage reader returns `null` (no session) when JSON parsing fails, rather than returning the raw corrupt string. Prevents downstream crashes from treating garbage as a session object.

```typescript
// OLD: returns raw garbage string on parse failure
// NEW: returns null — treat corrupted entry as absent
```

## Implementation Guidance (Kotlin)

```kotlin
fun readSessionFromStorage(): UserSession? {
    val raw = storage.retrieve(storageKey) ?: return null
    return try {
        Json.decodeFromString<UserSession>(raw)
    } catch (e: SerializationException) {
        null  // Corrupted — treat as no session
    }
}
```

### Key Behaviors to Match
- Malformed session JSON → return null, not throw
- SDK gracefully prompts re-auth instead of crashing

## Acceptance Criteria
- [ ] Malformed session JSON returns null, not throws
- [ ] Test verifies corrupted storage returns null session

## Context
- **supabase-js version**: v2.106.0
- **Related**: SDK-1038 (dart), SDK-1039 (py), SDK-1040 (swift)
- **Parity tracking**: Auto-generated by `/sync-sdk-parity`

### LINKED ISSUE #1318: auth: loadFromStorage logs "Failed to load session" at ERROR for the normal no-session case (e.g. cold start while logged out)

PierreVieira:
## Summary

Closes #1318 and #1314.

`AuthImpl.loadFromStorage()` logged **`Failed to load session` at ERROR with a full stacktrace** whenever there was no stored session — which is a normal, expected state (any logged-out user, on every cold start, on all platforms, and on Android also after `signOut()` + foreground via the `onStart` lifecycle callback). It was caught (no crash, `loadFromStorage` returns `false`), but the ERROR-level log + stacktrace was misleading: it looked like something failed when nothing did.

### Root cause

- `SettingsSessionManager.loadSession()` and `MemorySessionManager.loadSession()` threw a plain `IllegalStateException` via `error(...)` when there was simply no stored session.
- `AuthImpl.loadFromStorage()` logged **any** caught exception at ERROR, so the expected empty-session case was logged identically to a genuine failure (e.g. a stored session that fails to deserialize).

### Fix

1. Introduce a dedicated `NoSessionFoundException` thrown by both `SettingsSessionManager` and `MemorySessionManager` when no session is stored. It **extends `IllegalStateException`** so existing code that catches that type keeps working (backwards-compatible).
2. In `AuthImpl.loadFromStorage()`, catch `NoSessionFoundException` separately and log it at **debug** (`"No session found in storage"`), while still logging any other exception at **error**. Genuine failures (e.g. deserialization errors) are still surfaced at ERROR.

The change is minimal and backwards-compatible — `loadSessionOrNull()` keeps its default implementation, so custom `SessionManager` implementations don't need to change.

### Why a dedicated exception instead of just switching to `loadSessionOrNull()`

Switching `loadFromStorage()` to `loadSessionOrNull()` would collapse *all* failures — including a corrupt/undeserializable stored session — into a silent `null`, losing the ERROR log for real problems. The dedicated exception keeps that distinction: **no session → debug**, **corrupt session → error**, both treated as "not logged in" without crashing.

### #1314 (parity: return null on JSON parse failure)

A stored session that fails to deserialize is treated as absent: `SessionManager.loadSessionOrNull()` returns `null` (so it never crashes the load path), while `loadFromStorage()` still logs it at ERROR for observability — a corrupt entry is a real problem, unlike the normal no-session case. Happy to silence it entirely instead if you'd prefer strict supabase-js parity.

## Tests

- `MemorySessionManagerTest` / `SettingsSessionManagerTest` — assert `loadSession()` throws `NoSessionFoundException` when empty.
- `SettingsSessionManagerTest.testLoadSessionOrNullReturnsNullOnCorruptSession` — a corrupt stored session is treated as absent (`loadSessionOrNull()` returns `null`), covering #1314.

`./gradlew :Auth:jvmTest` and `./gradlew :Auth:detektMetadataMain` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jan-tennert:
While its a good approach, I don't think we should introduce logging tests in this PR. Otherwise looks good

PierreVieira:
> While its a good approach, I don't think we should introduce logging tests in this PR. Otherwise looks good

Ok. Dropped the logging-based test in b1e2066. Kept the behavioral ones (typed `NoSessionFoundException` when empty, `null` on a corrupt session). Thanks for the review!

jan-tennert (approved):

PierreVieira:
Heads up: the failing **Integration Tests** check is unrelated to this PR. It fails on all 6 `PostgrestIntegrationTest` tests (`PostgrestRestException`), and the same failure is present on `master` ([run](https://github.com/supabase-community/supabase-kt/actions/runs/27652558836)) — so it's a pre-existing CI/infra issue, not caused by this change. Every other check here is green (`jvmTest` incl. the new Auth tests, all native targets, `detekt`, sample builds).

Tracked in #1332.

jan-tennert (approved):

PR #1322, opened 2026-06-14
feat(auth): expose public getter for AuthConfig.httpCallbackConfig on Desktop

### LINKED ISSUE #1321: [Feature request]: Public getter for AuthConfig.httpCallbackConfig (Desktop) to update redirectHtml at runtime without reflection
### General Info

- supabase-kt version: 3.6.0 (also verified on current `master`, commit 87601ee)
- Platform: Desktop (JVM) OAuth http callback

### Feature request

Expose a **public getter** for `AuthConfig.httpCallbackConfig` on Desktop so the active `HttpCallbackConfig` (and its already-public `redirectHtml`) can be read/mutated at runtime from outside the `io.github.jan.supabase.auth` module, without reflection.

Today, in `Auth/src/desktopMain/.../AuthConfig.kt`, the field is `internal`:

```kotlin
actual class AuthConfig : AuthConfigDefaults() {
    internal var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()

    fun httpCallbackConfig(block: HttpCallbackConfig.() -> Unit) {
        httpCallbackConfig = HttpCallbackConfig().apply(block)
    }
}
```

The builder `httpCallbackConfig { ... }` only runs **once at client install time**. There is no public API to read the current `httpCallbackConfig` instance afterwards. Note that `HttpCallbackConfig.redirectHtml` is itself a **public `var`** — the only thing blocking a no-reflection update is the `internal` visibility of the `httpCallbackConfig` field.

#### Proposed change (additive, backwards-compatible)

Make the getter public while keeping the setter `internal`:

```kotlin
actual class AuthConfig : AuthConfigDefaults() {
    var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()
        internal set

    fun httpCallbackConfig(block: HttpCallbackConfig.() -> Unit) {
        httpCallbackConfig = HttpCallbackConfig().apply(block)
    }
}
```

This changes only the getter's visibility (`internal` → public); the setter stays `internal`, so the configured-once-via-builder contract is preserved and no existing code breaks.

### Usecase

On Desktop, the OAuth http callback route reads `auth.config.httpCallbackConfig.redirectHtml` fresh on every request (see `HttpCallbackRoutes.kt`). We want to **swap the redirect/success HTML page between sign-in attempts at runtime** — e.g. localized "you may close this window" pages depending on the provider or the app's current language — which the install-time builder can't express.

Because the field is `internal`, our only option today is reflection, which is brittle across versions:

```kotlin
// ApplySupabaseRedirectHtmlUseCase.kt — current workaround
operator fun invoke(html: String) {
    val authConfig = supabaseClient.auth.config
    val httpConfigField = authConfig.javaClass
        .getDeclaredField("httpCallbackConfig")
        .apply { isAccessible = true }
    val httpConfig = httpConfigField.get(authConfig)
    httpConfig.javaClass
        .getDeclaredField("redirectHtml")
        .apply { isAccessible = true }
        .set(httpConfig, html)
}
```

With the proposed getter, this collapses to a single line against the public API:

```kotlin
operator fun invoke(html: String) {
    supabaseClient.auth.config.httpCallbackConfig.redirectHtml = html
}
```

Real-world example (our project): [ApplySupabaseRedirectHtmlUseCase.kt](https://github.com/Quartel-Enterprise/bible-planner-mobile-client/blob/8036635d3883c09de3d44c082b444f5f6baae3e1/feature/login/src/jvmMain/kotlin/com/quare/bibleplanner/feature/login/domain/usecase/ApplySupabaseRedirectHtmlUseCase.kt)

I'm happy to open a PR for the getter change if this looks good.

PierreVieira:
## What

Makes the Desktop `AuthConfig.httpCallbackConfig` getter **public** while keeping the setter `internal`.

```kotlin
// Auth/src/desktopMain/.../AuthConfig.kt
var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()
    internal set
```

## Why

The field was fully `internal`, so the active `HttpCallbackConfig` could only be configured **once at client install time** via the `httpCallbackConfig { }` builder, and never read or mutated afterwards from outside the auth module.

`HttpCallbackConfig.redirectHtml` is already a public `var` and is re-read on **every** OAuth callback request (`HttpCallbackRoutes.kt`), so it's meant to be updatable at runtime — but the only way to reach the instance from app code today is reflection.

Exposing the getter (setter stays `internal`) lets callers do:

```kotlin
supabaseClient.auth.config.httpCallbackConfig.redirectHtml = newHtml
```

instead of the brittle reflection workaround. The configure-once-via-builder contract is preserved (setter remains internal), and the change is **purely additive / backwards-compatible** — no public symbol is removed or changed, only a getter's visibility is widened.

## Use case

We swap the desktop OAuth success/redirect HTML between sign-in attempts (localized "you may close this window" pages depending on app language). The install-time builder can't express this. Current reflection-based workaround in our project: [ApplySupabaseRedirectHtmlUseCase.kt](https://github.com/Quartel-Enterprise/bible-planner-mobile-client/blob/8036635d3883c09de3d44c082b444f5f6baae3e1/feature/login/src/jvmMain/kotlin/com/quare/bibleplanner/feature/login/domain/usecase/ApplySupabaseRedirectHtmlUseCase.kt).

## Testing

`./gradlew :auth-kt:compileKotlinJvm` succeeds.

Closes #1321


Auth/src/desktopMain/kotlin/io/github/jan/supabase/auth/AuthConfig.kt
```diff
@@ -9,7 +9,15 @@ import kotlin.time.Duration.Companion.minutes
...
-    internal var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()
+    /**
+     * The current http callback configuration for the web server, when logging in with OAuth or SSO.
+     *
+     * The setter is internal; configure it via [httpCallbackConfig]. The getter is public so the
+     * active [HttpCallbackConfig] (e.g. [HttpCallbackConfig.redirectHtml]) can be read or mutated at
+     * runtime without reflection.
+     */
+    var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()
+        internal set
```
jan-tennert, line 16:
```suggestion
    /**
     * The current http callback configuration for the web server, when logging in with OAuth or SSO.
     */
    var httpCallbackConfig: HttpCallbackConfig = HttpCallbackConfig()
        internal set
```
I think a better solution would be to just make the config overridable per request via the ExternalAuthConfig, because the plugin's config is not supposed to be mutated, however I'm freezing breaking changes to Auth until #1291 is merged, so that will have to do.

PierreVieira: Thanks, applied your suggestion 👍

And that makes sense — a per-request override via `ExternalAuthConfig` would be the cleaner long-term shape (no mutation of the plugin's shared config). Happy to follow up with that once the Auth freeze lifts after #1291. For now this interim getter unblocks updating `redirectHtml` at runtime without reflection.

jan-tennert (approved):

PR #1326, opened 2026-06-15
Add support for Realtime Protocol V2

jan-tennert:

Feature

Only V1 is supported (JSON payloads)

## What is the new behavior?

V2 is also supported and the default (binary broadcast support)

Syntax:
TODO


jan-tennert:
merged into #1325

Issue #1115, opened 2025-12-08
[Feature request]: Add tests for custom exception types


jan-tennert:
### General Info

- [x] I installed the latest version of supabase-kt
- [x] I checked for similar feature requests

### Feature request

`RestExceptions` and their plugin specific subtypes

PR #1368, opened 2026-07-26
Add tests for RestException and plugin-specific subtypes

MohammedThalha2002:
#1115  Covers the RestException base class and its four built-in subclasses, plus the Auth, Postgrest, Storage and Function parseErrorResponse implementations.

- new RestExceptionTest for the composed message, statusCode and header/URL masking
- new StorageExceptionTest and FunctionsExceptionTest (previously uncovered)
- six additional AuthRestException cases (error code mapping, session_not_found,
  401/422/other fallbacks)
- one PostgrestRestException case for its Code/Hint/Details description

### DIFF
--- a/Auth/src/commonTest/kotlin/AuthRestExceptionTest.kt
+++ b/Auth/src/commonTest/kotlin/AuthRestExceptionTest.kt
@@ -2,16 +2,21 @@ import io.github.jan.supabase.SupabaseClient
 import io.github.jan.supabase.SupabaseClientBuilder
 import io.github.jan.supabase.auth.Auth
 import io.github.jan.supabase.auth.auth
+import io.github.jan.supabase.auth.exception.AuthErrorCode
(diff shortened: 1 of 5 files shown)


jan-tennert (approved):

MohammedThalha2002:
CI's test (jvm) failed on RealtimeChannelTest > testChannelStatusWithoutPostgres, which is unrelated to this PR. It only touches commonTest in Supabase/Auth/Postgrest/Storage/Functions, and Realtime isn't in the diff. It looks like a pre-existing race: unsubscribe() sets UNSUBSCRIBING and then suspends sending the LEAVE frame, so the mock ws reply can flip the status to UNSUBSCRIBED before the next assertion reads channel.status.value. The Realtime plugin runs on Dispatchers.IO while the test body runs on runTest's test dispatcher, so the interleaving is scheduler-dependent.

PR #1391, opened 2026-09-06
fix(auth): [Security] Avoid writing session credentials to the platform log & mask sensitive values

hieuwu:
Stop writing session credentials to the platform log

UserSession is a data class holding four bearer credentials as plain Strings, so the compiler-synthesized toString() prints all four verbatim. Four call sites interpolated credential-bearing values straight into log lines:

| # | Location | Leaked value |
|---|---|---|
| 1 | **AuthImpl.kt:569** `logger.d { "Importing session $session from $source, ..." }` | `accessToken`, `refreshToken`, `providerToken`, `providerRefreshToken` via `$session` `$source`. As `SessionSource.Refresh` / `UserChanged` / `UserIdentitiesChanged`  wrap a `UserSession` |
| 2 | **AuthExtensions.kt:27** `logger.d { "Fragment parts: $sessionParts" }` | The parsed OAuth redirect fragment map, which by construction holds `access_token` and `refresh_token` |
| 3 | **url/HashParsing.kt:9** `logger.d { "Parsing fragment/hash $hash" }` | The **raw** OAuth redirect fragment, credentials included |
| 4 | **Utils.kt:30** `logger.e(e) { "Couldn't retrieve user using access token $accessToken..." }` | A raw access token at **`ERROR`**, which is enabled under the shipped default log level |

## What is the new behavior?

Fixed at the **type** as root, not at the call sites. New data class that if wraps a `UserSession` will not redintroduce the leak

**1. Mask at the type boundary.** `UserSession` overrides `toString()` to render every credential through `StringMasking.maskString`, keeping a two-character prefix and the length so the value stays debuggable.

 `SessionSource.Refresh` / `UserChanged` / `UserIdentitiesChanged` are data classes wrapping a `UserSession`, their `toString()` delegates to the overridden one

**2. Mask the OAuth redirect fragment.** Not related to `UserSession`,  two new helpers in `StringMasking` to cover more sensitive parameters
 - `maskParameters(Map<String, String>)` for a parsed map
 - `maskParameterString(String)` for a raw `a=b&c=d` fragment or query string. 
Both mask a denylist of credential parameter names (`access_token`, `refresh_token`, `provider_token`, `provider_refresh_token`, `id_token`, `code`, `code_verifier`) and leave everything else readable, so `expires_in=3600&token_type=bearer` still shows. 
`maskParameterString` preserves ordering and passes malformed segments through unchanged, so the output stays useful for diagnosing a bad redirect.

**3. Header masking is now case-insensitive and wider.** `SENSITIVE_HEADERS` became a lowercase `Set` matched against `key.lowercase()`, extended with `cookie`, `set-cookie`, `proxy-authorization` and `x-api-key`. Bearer-prefix handling is driven by the value (`startsWith("Bearer ")`) rather than an exact-case key comparison, and every value of a multi-valued header is masked rather than just the first.

## Additional context

**References and practices.**
- **Domain primitives**: a value with a security invariant enforces that invariant itself rather than delegating it to every consumer, and `toString()` is the exact seam where it belongs, because that is what string interpolation calls. Johnsson, Deogun & Sawano, *Secure by Design* (Manning, 2019), Ch. 5; Ch. 9 ("Handling failures securely") makes the narrower point that log and exception output must not carry the secret that caused the failure.

- **RFC 6749 §10.3** — access token credentials must be kept confidential in transit *and
  storage*, and the section explicitly notes that the implicit grant exposes the token in the
  URI fragment. A platform log is storage; this is the control the four call sites broke.
  <https://www.rfc-editor.org/rfc/rfc6749#section-10.3>
- **OWASP MASVS-STORAGE-2** "The app prevents leakage of sensitive data" (OWASP MASVS v2),
  naming logs as an unintentional leakage sink, with **MASWE-0005: Insertion of Sensitive Data
  into Logs** as the mapped weakness.
  <https://mas.owasp.org/MASVS/controls/MASVS-STORAGE-2/>
- **OWASP ASVS 4.0.3 §7.1.1** the application must not log credentials; ASVS maps this
  requirement to CWE-532. Masking here reduces exposure consistent with the intent of 7.1.1;
  the control's own remedy for session tokens is an irreversible hashed form, which a
  prefix-plus-length rendering is not.
- **CWE-532** Insertion of Sensitive Information into Log File.
  <https://cwe.mitre.org/data/definitions/532.html>

- **RFC 6749 §4.2.2** defines the implicit-grant redirect fragment and its parameters
  (`access_token`, `token_type`, `expires_in`, `refresh_token`, `state`), delivered in
  `application/x-www-form-urlencoded` form. This is the parameter set `SENSITIVE_PARAMETERS`
  is drawn from. <https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2>


Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/user/UserSession.kt
```diff
@@ -25,4 +27,50 @@ data class UserSession(
...
     val type: String = "",
     val expiresAt: Instant = Clock.System.now() + (expiresIn.seconds),
-)
\ No newline at end of file
+) {
+
+    /**
+     * Renders every field, printing each bearer credential through [renderToken].
+     *
+     * This is the single place that knows the field list, so [toString] and [unsafeToString] can
```
jan-tennert: On this and the other methods, the docs might be **too** detailed. I mean it doesn't hurt, but I don't think the KDocs is a good place to "justify" why a method exists like it does. Documenting what it does should be enough. Same for the methods below

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/Utils.kt
```diff
@@ -27,6 +28,8 @@ internal suspend fun Auth.tryToGetUser(accessToken: String) = try {
     retrieveUser(accessToken)
 } catch (e: Exception) {
     currentCoroutineContext().ensureActive()
-    logger.e(e) { "Couldn't retrieve user using access token $accessToken.\nIf you use the project secret, ignore this message." }
+    // Masked: this is logged at ERROR, which is enabled under the default LogLevel.INFO,
+    // so an unmasked token here reaches the platform log in a stock configuration.
```
jan-tennert: Not sure if explaining this in this specific instance is worth it (because the token should be masked even with the debug level)

Auth/src/commonTest/kotlin/SessionCredentialLoggingTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    SECRET_PROVIDER_TOKEN,
+    SECRET_PROVIDER_REFRESH_TOKEN,
+)
+
+/**
+ * Auth logs from a background coroutine during `init`, so the recorder has to tolerate concurrent
+ * appends while the test reads. An immutable list behind a CAS loop keeps reads snapshot-consistent.
+ */
+@OptIn(ExperimentalAtomicApi::class)
+private class RecordingLoggingProcessor : SupabaseLoggingProcessor {
```
jan-tennert: Same thing like in a different PR: I don't think introducing logging test infrastructure for a single feature test makes sense. I think we should commonize that, so that ideally we can test the logs across all modules, but that wouldn't fit in this PR I think. Or we atleast have to think about how we make that useful for all modules and not only for one feature test

hieuwu: I thought the same, just try to verify the log, seems unnecessary. 
There's a way that we can have static code analysis with SonarQube (https://www.sonarsource.com/solutions/security/sca/) , which can helps spot these issues across modules early.
They also offer free usage for Open Source with public report. Let me try some set up and see if it's suitable

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/Utils.kt
hieuwu: Just removed, it's redundant

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/user/UserSession.kt
hieuwu: Yeah make sense.

Auth/src/commonMain/kotlin/io/github/jan/supabase/auth/AccessToken.kt
```diff
@@ -60,7 +59,7 @@ private suspend fun SupabaseClient.checkAccessToken(token: String) {
                 Auto refresh running: $autoRefreshRunning
                 Auto-refresher information: ${auth.autoRefreshInformation()}
                 OS: ${OSInformation.CURRENT}
-                Session: ${StringMasking.maskSession(currentSession)}
+                Session: $currentSession
```
jan-tennert, line 62: Was that intentional?

hieuwu: Yes. `currentSession` here is UserSession which has toString with masked value. So when call `$currentSession` in a string it will use toString function to get the object string with masked values. 

Screenshot 2026-09-10 at 21 17 31

jan-tennert (approved):

PR #1395, opened 2026-09-10
fix(realtime): prevent heartbeat tick from racing disconnect()

### LINKED ISSUE #1394: [Bug]: Realtime heartbeat tick races disconnect() and crashes the app with "Websocket not yet initialized"
## General Info
- [x] I checked the issues for similar bug reports
- [x] I installed the latest version of supabase-kt
- [x] I checked the troubleshooting page for similar problems

## Version(s)
3.7.0 in production; reproduced on 3.8.0 and current `master` (615d3ef)

## Kotlin Version
Kotlin 2.4.10

## Kotlin Target(s) and their respective versions
Android 11–16 (crash observed in production), reproduced on JVM 17

## What happened? (include your code)

`RealtimeImpl.startHeartbeating()` launches a child coroutine per tick that calls `sendHeartbeat()`, which dereferences the `websocket` property. That getter throws `IllegalStateException("Websocket not yet initialized")` when `_websocket` is null.

`disconnect()` clears `_websocket` **before** cancelling `heartbeatJob`, and cancellation is cooperative anyway, so a tick that is already running on another worker thread reaches the getter after the reference was cleared. The realtime `CoroutineScope` is `dispatcher + SupervisorJob()` with no `CoroutineExceptionHandler`, so the exception escapes to the thread's uncaught-exception handler. On Android that terminates the process.

`disconnect()` is reached from three normal paths, all of which can coincide with a heartbeat tick: `reconnect()` after a socket error, the `SessionStatus.NotAuthenticated` listener (`disconnectOnSessionLoss`), and the deferred disconnect after the last channel is removed. On Android this happens most often when a backgrounded/frozen process is thawed: the heartbeat timer and the socket error fire in the same instant.

Production stack trace (Crashlytics, app in background, 3.7.0):

```
java.lang.IllegalStateException: Websocket not yet initialized
    at io.github.jan.supabase.realtime.RealtimeImpl.getWebsocket(RealtimeImpl.kt:58)
    at io.github.jan.supabase.realtime.RealtimeImpl.sendHeartbeat(RealtimeImpl.kt:263)
    at io.github.jan.supabase.realtime.RealtimeImpl.access$sendHeartbeat(RealtimeImpl.java:52)
    at io.github.jan.supabase.realtime.RealtimeImpl$startHeartbeating$1$1.invokeSuspend(RealtimeImpl.java:198)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
    ...
```

Expected: a heartbeat tick that overlaps with `disconnect()` is skipped; nothing escapes the realtime scope.

The app never calls `disconnect()` itself; it only subscribes to `postgresChangeFlow` channels and removes them when screens close.

## Steps To Reproduce

A JVM test that uses a no-op `RealtimeWebsocketFactory`, `heartbeatInterval = 1.milliseconds`, `coroutineDispatcher = Dispatchers.Default`, and calls `realtime.disconnect()` / `realtime.connect()` in a loop from another thread while capturing `Thread.setDefaultUncaughtExceptionHandler` reproduces it in under a second on `master`:

```
RACE-TEST: escaped=IllegalStateException: Websocket not yet initialized after 895 ms
```

It needs real parallelism (`Dispatchers.Default`); a single-threaded test dispatcher cannot interleave inside `sendHeartbeat()`. I will open a PR with the test and a fix (cancel the jobs before clearing the reference, and make the heartbeat tick null-safe).

### LINKED ISSUE #1398: 'Websocket not yet initialized' crash
Supabase's realtime library (supabase-kt) keeps exactly one live object per client: a WebSocket connection, stored in a variable I'll call _websocket. It starts out empty (null). Three things happen to it:

- Connect: creates the socket, stores it in _websocket, then flips a separate "status" flag to CONNECTED.
- Disconnect: clears _websocket back to empty, then flips the status flag to DISCONNECTED.
- Anything that wants to use the socket (send a message, ping, rejoin a channel) reads _websocket — and if it's empty, it throws: "Websocket not yet initialized". That crash you're seeing is always that exact one line of code inside the library.

Why it crashes even though it looks "connected"

The library also runs two background loops on its own, entirely inside itself, that we (the app) never call directly:

1. A heartbeat, every 15 seconds, to keep the connection alive.
2. A rejoin-all-channels step, right after every reconnect, to re-subscribe everything that was subscribed before.

Both of these were scheduled at some earlier point while the socket was up. The bug: between the moment they were scheduled and the moment they actually run, the socket can drop — a network blip, the phone's radio going into Doze/background throttling, a server-side timeout, anything ordinary on a mobile connection. When that happens, disconnect() clears _websocket — but the heartbeat tick or the rejoin step that was already queued up doesn't know that happened, and runs anyway, reads the now-empty _websocket, and throws.
```

Fatal Exception: java.lang.IllegalStateException: Websocket not yet initialized
       at io.github.jan.supabase.realtime.RealtimeImpl.getWebsocket(RealtimeImpl.kt:58)
       at io.github.jan.supabase.realtime.RealtimeChannelImpl.updateAuth(RealtimeChannelImpl.kt:203)
       at io.github.jan.supabase.realtime.RealtimeImpl$setAuth$3.invokeSuspend(RealtimeImpl.kt:251)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
       at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.java:124)
       at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.java:586)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:798)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

alperen-kaplan:

Bug fix

Fixes #1394

A heartbeat tick launched by `startHeartbeating()` can still be running when `disconnect()` clears `_websocket`. `sendHeartbeat()` then hits the throwing `websocket` getter (`"Websocket not yet initialized"`). The realtime scope has no `CoroutineExceptionHandler`, so the exception escapes to the thread's uncaught-exception handler; on Android that kills the process. Observed in production (Crashlytics, 3.7.0) and reproduced on `master`.

## What is the new behavior?

Two small changes in `RealtimeImpl`:

1. `disconnect()` cancels `messageJob` and `heartbeatJob` **before** clearing the websocket reference, so new ticks cannot observe a null websocket.
2. `sendHeartbeat()` reads `_websocket.load()` and skips the tick (debug log) when it is null. This covers the remaining window, because cancellation is cooperative and a tick that is already running is not stopped by `cancel()`.

A JVM regression test (`RealtimeHeartbeatDisconnectRaceTest`) drives `disconnect()`/`connect()` from another thread with a 1 ms heartbeat on `Dispatchers.Default` and asserts that nothing reaches the uncaught-exception handler. It lives in `jvmTest` on purpose: the race needs real parallelism, which the single-threaded test dispatcher used by the `commonTest` suite cannot provide.

Results on this branch:

- unpatched `master`: the test fails after ~0.9 s with `IllegalStateException: Websocket not yet initialized`
- patched: 20 s of continuous disconnect/connect, nothing escapes
- `./gradlew :realtime-kt:jvmTest`: all existing Realtime tests pass (64) plus the new one

## Additional context

`disconnect()` is reached from three normal paths that can all coincide with a tick: `reconnect()` after a socket error, the `SessionStatus.NotAuthenticated` listener, and the deferred disconnect after the last channel is removed. On Android the race is most likely when a frozen background process is thawed and the heartbeat timer and the socket error fire together, which matches the production breadcrumbs (all crashes in the background).


Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt
```diff
@@ -203,10 +203,12 @@ import kotlin.time.Duration
 
     override fun disconnect() {
         logger.d { "Closing websocket connection" }
+        // Cancel the jobs that dereference the websocket before clearing it, so a heartbeat
```
jan-tennert: I think thats self-explanatory

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt
```diff
@@ -259,6 +261,12 @@ import kotlin.time.Duration
             reconnect()
             return
         }
+        // Cancellation is cooperative: a tick may still be running while disconnect() clears
+        // the websocket. Skip the tick instead of throwing into the scope (no handler there).
```
jan-tennert: that too

Realtime/src/jvmTest/kotlin/RealtimeHeartbeatDisconnectRaceTest.kt
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * `startHeartbeating()` launches a child coroutine per tick that dereferences `websocket`
+ * (which throws "Websocket not yet initialized" when `_websocket` is null). `disconnect()`
+ * stores null *before* cancelling `heartbeatJob`, so a tick that is already running on another
+ * worker thread hits the throwing getter. The scope has no CoroutineExceptionHandler, so the
+ * exception escapes to the thread's uncaught handler — on Android that kills the process.
+ *
+ * Runs on real threads (Dispatchers.Default) because the race needs parallelism; a
+ * single-threaded test dispatcher cannot interleave inside `sendHeartbeat()`.
+ */
```
jan-tennert:
```suggestion
 * Regression test for the heartbeat ↔ disconnect() race.
 *
 * `startHeartbeating()` launches a child coroutine per tick that dereferences `websocket`
 * (which throws "Websocket not yet initialized" when `_websocket` is null). `disconnect()`
 * stores null *before* cancelling `heartbeatJob`, so a tick that is already running on another
 * worker thread hits the throwing getter. The scope has no CoroutineExceptionHandler, so the
 * exception escapes to the thread's uncaught handler — on Android that kills the process.
 *
 * Runs on real threads (Dispatchers.Default) because the race needs parallelism; a
 * single-threaded test dispatcher cannot interleave inside `sendHeartbeat()`.
 */
```
That can be shortened. You don't need to justify code, just describe what it does. It can literally just be a one-liner:
**Catches potential regression, where sending heartbeats causes an uncaught exception because it was run after the websocket got cancelled**

jan-tennert (changes requested):

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt
alperen-kaplan: Removed in b16662f.

Realtime/src/commonMain/kotlin/io/github/jan/supabase/realtime/RealtimeImpl.kt
alperen-kaplan: Removed in b16662f.

Realtime/src/jvmTest/kotlin/RealtimeHeartbeatDisconnectRaceTest.kt
alperen-kaplan: Replaced with the one-liner in b16662f.

alperen-kaplan:
Thanks for the review! Dropped both inline comments and shortened the test KDoc to your one-liner (b16662f). No logic change; `:realtime-kt:jvmTest` is still green locally.

jan-tennert (approved):

PR #1397, opened 2026-09-15
Gate public ABI changes with Kotlin ABI validation

hieuwu:
Build/CI infrastructure + docs. No production source file changed,the only touch is Gradle config; every */api/*.api file is a new baseline, not an edit.

supabase-kt ships to Maven Central at 3.x with no mechanical check that its public ABI stays stable.
Detekt runs on every PR but doesn't model binary compatibility, so only a reviewer's eye stands between a contributor and a silent binary break.

Example of invisible failure without a dump:
Source-compatible changes that are binary-incompatible: SupabaseClientConfig / SupabaseLoggingConfig / SupabaseNetworkConfig are public data classes reachable from SupabaseClient.config. Adding one defaulted property compiles cleanly and changes nothing at source level, but produces this:

```diff
- public final fun copy (...;Lio/github/jan/supabase/OSInformation;)Lio/github/jan/supabase/SupabaseClientConfig;
- public static synthetic fun copy$default (...;Lio/github/jan/supabase/OSInformation;ILjava/lang/Object;)...
+ public final fun copy (...;Lio/github/jan/supabase/OSInformation;Z)Lio/github/jan/supabase/SupabaseClientConfig;
+ public static synthetic fun copy$default (...;Lio/github/jan/supabase/OSInformation;ZILjava/lang/Object;)...
+ public final fun component10 ()Z
```

The - lines are the break: every consumer that called copy() dies with NoSuchMethodError at runtime. A reviewer sees "optional field, defaulted" and approves. Nothing else in the build reports it.

Why this should land before other API work: without a recorded baseline there's no way to prove a refactor didn't break. The dump makes expand-contract enforceable

## What is the new behavior?

File | What it does
-- | --
buildSrc/.../AbiValidation.kt | Enables abiValidation per module. `keepLocallyUnsupportedTargets = true` so targets that can't build on the current host keep their recorded declarations instead of being deleted. Deferred via pluginManager.withPlugin, the script configures modules before they apply the KMP plugin.
build.gradle.kts | Applies it to published modules; registers `apiCheck` / `apiDump` aggregates mirroring the existing detektAll. They depend on task paths, since the per-module ABI tasks register lazily.
*/api/ | 22 baseline dumps, 8 modules, ~20k lines — Auth, Functions, Postgrest, Realtime, Storage, Supabase, serializers/Jackson, serializers/Moshi. One .api per JVM/Android target, one .klib.api covering all 15 Native/JS/Wasm targets.
.github/workflows/api-check.yml | The Check public ABI job.
CONTRIBUTING.md | How to read a dump diff and what SemVer requires in each case.

## Additional context
[Kotlin ABI validation](https://kotlinlang.org/docs/gradle-binary-compatibility-validation.html) 
[Kotlin backward-compatibility guidelines](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html)


jan-tennert:
Hi, thanks for this PR. This is definitely planned, but I think we should wait until we figured out the big refactor™ before making any big changes / implement ABI validation. Don't think that makes sense for this version as a lot of stuff is getting changed.

hieuwu:
> Hi, thanks for this PR. This is definitely planned, but I think we should wait until we figured out the big refactor™ before making any big changes / implement ABI validation. Don't think that makes sense for this version as a lot of stuff is getting changed.

No worries. I just raised here to see if it is needed. It should be done after the refactor.

